PR #30, opened 2026-06-23
feat: more efficient fake suspend

xXJSONDeruloXx:
Hello! I installed Armada to my RP6's internal storage and have been having a blast with it over the last few days. One thing I did notice is that because of SM8550 quirks, true suspend is not feasible/reliable without unpredictable consequences. I took a look at your fake-suspend implementation and upstream rocknix, then took a stab at refining the scope of its reach to reduce idle draw, and most importantly; safely cut down fan noise when it is safe to do so. 

The following changes here have been tested on my device across multiple suspended games, idle on Steam BPM home screen, and in desktop mode for good measure:

- freeze the complete `user.slice` instead of leaving graphical session and user services running. per proc freezer fallback if slice fails
- save state then block wifi and bluetooth, then ensure only those which were enabled are restored on wake
- gate fan activity at two tiers, one with a higher speed to rapidly cool temp down to acceptable level, and one other pwm speed till temp is below fan_off_below, at which point we can zero the fan, but still do the 3 sec probe to kick it up to one of those two speeds to keep things safe and cool (if nothing else, this makes the UX of sleeping feel a lot more "deck like", open to feedback on ranges or approach!)

can config those as needed in power-profile with current defaults of 

```
[suspend]
fan_off_below=45
fan_on_above=50
fan_safe_above=55
fan_low_pwm=51
fan_safe_pwm=128
```

There are fail safe behaviors too, if temp data can not be read, daemon will fall back to safe pwm.

on my RP6 after lots of testing the above changes, im seeing fake-suspend usage go down from around 1.79 watts to around 0.85 watts, and even while running heavy games where device is hot, suspending now runs the fan at aggressive pwm for a bit, then ramps down to low pwm, then goes to 0 within around a minute, even when in my pocket (I get most of my gaming in while walking the dog lol).

please let me know what you think! and if you would like to gate some of these changes behind flags or per device configs, I fully understand and am open to any necessary stewardship 

Also, feel free to mark RP6 in your readme as tested working!

Thanks again for making such an awesome project!

### DIFF
--- a/system_files/usr/bin/armada-power
+++ b/system_files/usr/bin/armada-power
@@ -79,6 +79,7 @@ def status():
         "fan_pwm",
         "fan_rpm",
         "temperature",
+        "suspended",
     ):
         value = data[key].unpack() if hasattr(data[key], "unpack") else data[key]
         print(f"{key}={value}")
--- a/system_files/usr/libexec/armada/armada-powerd
+++ b/system_files/usr/libexec/armada/armada-powerd
@@ -176,12 +176,15 @@ class ArmadaPower:
         self.underclock_config = {}
         self.fan_curves = {}
         self.fan_config = {}
+        self.suspend_config = {}
         self.default_profile = ""
         self.profile = ""
         self.gpu_level = "auto"
         self.manual_gpu_clock = 0
         self.smoothed_temp = 0.0
         self.last_pwm = 0
+        self.suspended = False
+        self.suspend_fan_running = False
         self.suspend_save = {}
         self.connection = None
         self.load_config()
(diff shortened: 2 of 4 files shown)


virtudude:
@xXJSONDeruloXx Thanks for digging in here and making improvements. I'm traveling this week so might take me a few days to review and test but I'll try to get it merged as soon as I can.

virtudude:
@xXJSONDeruloXx do you see a very long wifi reconnect times (like 30+ seconds) when you wake the device? I can't tell if it's my wifi setup or the actual chip on these devices taking forever to reinit after rfkill.

PR #34, opened 2026-06-27
feat(power): implement true s2idle suspend with improved fake-suspend fallback

the8055u:
## Description
This PR significantly improves the suspend behavior on SM8550/SM8650/SM8750 devices by introducing support for true `s2idle` suspend while maintaining and enhancing the legacy `fake-suspend` mechanism.

The current `fake-suspend` approach results in higher battery drain because the SoC never truly sleeps. Real suspend (`s2idle`) is now automatically used if the kernel supports it, falling back to an improved `fake-suspend` otherwise.

## Changes
* **Added `s2idle` support in `fake-suspend`:** Automatically detected via `/sys/power/mem_sleep`. When active, it properly freezes tasks and suspends drivers via `echo freeze > /sys/power/state`, allowing CPUs to enter hardware idle states.
* **Improved `fake-suspend` fallback:** Offlines all secondary CPU cores via hotplug (`echo 0 > /sys/devices/system/cpu/cpu*/online`) during sleep to further reduce battery drain when `s2idle` is unavailable.
* **Added configurable `suspend_mode`:** Added to `power-profiles.conf` with `auto` (default), `s2idle`, and `fake` options to allow users to force a specific behavior.
* **Updated documentation:** Rewrote the "Power button and sleep" section in `README.md` to explain the new behavior.

## Testing
* **Test on device:** Verify `cat /sys/power/mem_sleep` to check if `s2idle` is available.
* Press power button to suspend and resume. Check battery drain compared to the old fake-suspend implementation.
* If `s2idle` causes GPU hangs (Adreno 740), users can safely revert to the legacy path by setting `suspend_mode=fake` in `/etc/armada/power-profiles.conf`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,10 +157,26 @@ desktop. The **Bazaar** app store and the **Armada Installer**
 
 ### Power button and sleep
 
-Pressing the power button does a "fake suspend" (inspired by ROCKNIX) rather than
-real S3 sleep: it blanks the screen and freezes the session, and the same press
-wakes it. Because the device does not truly sleep, idle battery drain is higher
-than it would be with real suspend.
+Pressing the power button suspends the device. Armada automatically selects
+the best suspend mode for your hardware:
+
+- **s2idle (suspend-to-idle):** If the kernel supports it, the device enters a
+  true low-power state — CPUs enter hardware idle, drivers are suspended, and
+  wake is instant via the power button IRQ. Battery drain is significantly lower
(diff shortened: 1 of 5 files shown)


virtudude:
@the8055u Were you able to test this successfully? My experience has been that s2idle almost never wakes from sleep.

virtudude:
This was superseded by the Odin 3 mem sleep support

PR #80, opened 2026-07-08
Nested gaming session for dual-screen devices (Thor phase 2)

brycesub:
Phase 2 of AYN Thor dual-screen support (#28): gaming mode with **both screens live** — Steam on the top screen, a full plasma desktop on the bottom — and near-instant mode switching.

Builds on #78 (its two commits are included here; this PR reduces to the two `thor-nested-gaming` commits once #78 merges). Also needs virtudude/armada-packages#2 (gamescope `wl_touch` patch), or the top screen loses touch in nested mode.

## Architecture

With `ARMADA_GAMING_SESSION=nested` (set for Thor), the compositor stack is inverted: KWin stays the only DRM master driving both panels, and gaming mode runs as a **nested gamescope fullscreen window** on the primary screen (`/usr/libexec/armada/nested-gaming`).

- **Switching:** `steamos-manager`'s session-bus instance handles Steam's `SwitchToDesktopMode`/`SwitchToGameMode` by stopping/starting the nested gamescope — the desktop is already underneath, so switching is near-instant (no SDDM session swap, no logout). Single-screen devices and the system-bus instance keep the existing session-swap path.
- **Boot:** `armada-session-default` still resets to "gamemode" each boot; on nested devices that session is now `armada-plasma`, and `desktop-bootstrap` launches the nested gamescope on fresh session start. Boot lands in gaming mode with the desktop live on the bottom screen.
- The embedded gamescope session is no longer used on nested devices (it remains untouched for every other device).

## Validated on hardware (AYN Thor)

- Cold boot → Steam on top + desktop on bottom, automatically
- Instant switch in both directions through Steam's UI, repeatedly, plus stability across multiple days of use
- Game performance indistinguishable from the embedded session (user-tested)
- Brightness (via the #78 backlight steering), gamepad, and per-screen touch all work; the bottom touchscreen drives the desktop while gaming

## Refresh switching

Steam's per-game refresh switching works through `nested-refresh-bridge`: it watches the `GAMESCOPE_DYNAMIC_REFRESH` atom Steam sets on the nested Xwayland root and applies the nearest supported panel mode via kscreen-doctor (restoring the max rate when dynamic refresh is off or the session ends).

Design doc: `docs/superpowers/specs/` (phase 2) — happy to include it in the PR if wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XMbeBSJJG1pNfdqkUcG1gn

PR #90, opened 2026-07-13
Add AYN button toggle for a bottom-screen performance HUD

brycesub:
## Summary

- add an AYN-button listener that swaps MangoApp between the nested top-screen overlay and a full-stat bottom-screen HUD
- pin the desktop MangoApp to the Thor's DSI-1 panel, force fullscreen behavior, and prevent it from stealing focus
- merge packaged KWin rules into the active per-user rulebook before nested gaming starts
- bind both HUD services to nested gaming and include VRAM in the bottom-screen statistics

Depends on #80, which provides the nested-gaming session this HUD integrates with.

## Validation

- all `tests/*.sh`
- ShellCheck on the HUD, nested-gaming, bootstrap, and test scripts
- `just check`
- device-side `systemd-analyze --user verify` for both HUD units
- live AYN Thor validation: bottom window measured `1240x1080` at the DSI-1 offset, remained fullscreen/above/skip-taskbar, and did not become the active window


tslmy:
@brycesub / other maintainers: shall we tag this as "hardware: Thor"?

JPyke3:
I do apologise but our dual screen implementation has heavily deviated from this original PR. As such I need to close it. More than happy to have this feature reintroduced with the new leasing approach if it works though!

PR #95, opened 2026-07-14
Add optional Nested Desktop

Janleyx:
## Summary

Add an opt-in SteamOS-style Nested Desktop for brief Plasma access from Gaming Mode without replacing Armada’s normal KWin Desktop Mode.

This revision addresses review feedback by limiting PR #95 to Nested Desktop and its necessary support:

- launch fullscreen KWin under the current Gamescope display path;
- isolate the nested runtime, D-Bus session, KScreen state, and KWin output state;
- share the parent PipeWire sockets;
- clear Gaming Mode’s forced Qt/X11 backend so Plasma creates a native Wayland panel;
- preserve normal landscape orientation at 150% logical scaling without forcing render resolution; and
- skip Armada’s physical-panel bootstrap inside the nested session.

`Armada Native` remains in this PR because Odin testing found that the ARM Steam client continued assigning default Proton when compatibility was unchecked. The explicit native compatibility entry is therefore required for the Nested Desktop shortcut to launch reliably as Linux-native.

## Split work

- Add to Steam: #236
- Desktop Mode Steam autostart: #237
- User-facing README documentation was removed as requested.

## Validation

Validated previously through reboots and repeated Gaming Mode → Nested Desktop → Desktop Mode cycles on an AYN Odin 2 Pro running Armada beta `20260729.d812101`. Nested Desktop remained landscape at 1920×1080 and 150% scale, while full Desktop Mode retained its independent physical-panel rotation.

For this rebased revision:

- relevant Bash launchers pass `bash -n`;
- the Nested Desktop application entry passes `desktop-file-validate`;
- `git diff --check` passes; and
- Armada’s full shell test suite passes on current upstream `main`.

### DIFF
--- a/system_files/usr/share/steam/compatibilitytools.d/armada-native/armada-native
+++ b/system_files/usr/share/steam/compatibilitytools.d/armada-native/armada-native
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -euo pipefail
+
+verb=${1:-}
+[[ -n "$verb" ]] || exit 2
+shift
+
+case "$verb" in
+    run|waitforexitandrun)
+        exec "$@"
+        ;;
+    destroyprefix)
+        exit 0
+        ;;
+    *)
+        echo "Unsupported Armada Native verb: $verb" >&2
+        exit 2
+        ;;
+esac
--- a/system_files/usr/bin/steamos-nested-desktop
+++ b/system_files/usr/bin/steamos-nested-desktop
@@ -1,0 +1,129 @@
+#!/usr/bin/bash
+# SPDX-License-Identifier: GPL-2.0-or-later
+set -euo pipefail
+
+if [[ ${XDG_SESSION_DESKTOP:-} == KDE || ${KDE_FULL_SESSION:-} == true ]]; then
+    echo "steamos-nested-desktop: refusing to launch inside KDE" >&2
+    if [[ ! -t 1 ]] && command -v kdialog >/dev/null 2>&1; then
+        kdialog --title "Nested Desktop" --sorry \
+            "Nested Desktop should be launched from Gaming Mode."
+    fi
+    exit 0
+fi
+
+state_dir="${XDG_STATE_HOME:-${HOME}/.local/state}/armada"
+install -d -m 0700 "${state_dir}"
+exec >>"${state_dir}/nested-desktop.log" 2>&1
+echo "=== Nested Desktop $(date --iso-8601=seconds) ==="
+
+# Gamescope's overlay, session-specific portals, and forced Qt backend must not
+# leak into Plasma. Armada exports QT_QPA_PLATFORM=xcb in Gaming Mode, which
+# prevents Plasma Shell from creating its Wayland panel surface.
+unset LD_PRELOAD XDG_DESKTOP_PORTAL_DIR QT_QPA_PLATFORM
+cd "${HOME}"
+
+parent_runtime=${XDG_RUNTIME_DIR:?XDG_RUNTIME_DIR is not set}
+nested_runtime=$(mktemp -d "${parent_runtime}/nested-desktop.XXXXXXXX")
+
+# The physical DSI panel and Gamescope's nested X11 output can expose the same
+# display identity even though they need different rotations. Mirror the user's
+# data/config directories with symlinks, but give Nested Desktop private legacy
+# KScreen and modern KWin output stores so its landscape transform cannot turn
+# Desktop Mode portrait.
+real_data_home=${XDG_DATA_HOME:-${HOME}/.local/share}
+nested_data_home="${nested_runtime}/data"
+install -d -m 0700 "${nested_data_home}/kscreen"
+shopt -s nullglob dotglob
+for entry in "${real_data_home}"/*; do
+    [[ ${entry##*/} == kscreen ]] && continue
+    ln -s "${entry}" "${nested_data_home}/${entry##*/}"
+done
+shopt -u nullglob dotglob
+
+real_config_home=${XDG_CONFIG_HOME:-${HOME}/.config}
+nested_config_home="${nested_runtime}/config"
+install -d -m 0700 "${nested_config_home}"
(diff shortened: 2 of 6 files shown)


virtudude:
@Janleyx Thanks for the contribution! Can you tell me more about the purpose of the nested desktop? One thing that concerns me about it is that KWin is actually a more efficient compositor than gamescope on this hardware, so it might regress performance in desktop mode. 

Also wondering if we still need the SD changes you have in here after I added jupiter-hw-support?

Janleyx:
oh hi! om so pleased you reached out! first off love the work behind the project and have been working hard on trying to help along. The whole point of nested desktop is to have quick access to your desktop enviroment from within gamemode so you dont have to switch to desktop mode to then switch back essentially. I made it very performant too. So in professional lingo: The purpose is to access Plasma briefly without switching sessions while retaining Steam Input, QAM, and the Steam overlay.

You’re right that this means KWin is running underneath Gamescope, so it can be less efficient than the direct KWin Desktop Mode session. I’m not claiming it improves performance. It tested acceptably after correcting the resolution/scaling configuration, but I did not run formal compositor benchmarks. Normal Desktop Mode remains the better choice for performance and extended use.

The nested launcher has no runtime effect unless the user explicitly adds and launches it. That said, I’m happy to split Nested Desktop into a separate PR—or remove it from this one—if you would prefer the desktop integration changes independently.

Regarding the SD changes: I agree that the Armada-side checks are no longer necessary now that I have the complete history.

The failure was reproduced on the June 28 image, which predated your July 3 jupiter-hw-support integration. That older image was missing the formatter, automounter, polkit integration, and f3. It also ran the older kernel without CONFIG_UNICODE.

The current source already has the primary native fixes:

jupiter-hw-support is installed in the image.

The kernel package has CONFIG_UNICODE=y.

I can remove the SD build assertions from this PR. I submitted the remaining defense-in-depth formatter change separately in virtudude/armada-packages#3, where the Jupiter RPM is maintained. It makes -O casefold conditional on the running kernel’s advertised support and was successfully built as a complete RPM on the Odin 2.

system_files/usr/share/steam/compatibilitytools.d/armada-native/armada-native
JPyke3, line 1: Is this one strictly necessary? I believe if you don't check "Force use of a steam Compatibility Tool", this will effectively do the same thing? 

If there is a reason why we need this in the compat tool list I am all ears, just not sure *why*

system_files/usr/libexec/armada/desktop-steam-autostart
JPyke3, line 1: I feel this one maybe deserves a seperate PR. It doesn't quite live in the scope of Nested Desktop.

I think it's something we do need to consider - we've talked about this a bit. But would prefer to have a seperate forum for it.

system_files/usr/bin/steamos-add-to-steam
JPyke3, line 1: This one may need it's own PR. I actually am a huge fan of the idea - It's just quite difficult to talk about so many features in one PR and have them all merged and tested together, not sure if you could split this one out?

README.md
JPyke3, line 1: I don't believe this needs to live in the README. If merged we can track on issue @ https://github.com/armada-os/armadaos.dev

JPyke3 (changes requested):
I am a big fan of alot of the changes in this PR! I think we just need to keep the scope of the PRs to one feature at a time as it helps us to validate alot of these changes alot easier

system_files/usr/share/steam/compatibilitytools.d/armada-native/armada-native
Janleyx: I added it as it didnt work to just disable the compaat tool. It uses the default proton by default but the armada native forces the os to run in linux the app.

system_files/usr/libexec/armada/desktop-steam-autostart
Janleyx: I agree it was my bad, i did this patch so that when you load up desktop mode you have controller support on desktop, another thing thta should be pre patched would be the desktop controls themselves which i had to manually do for the desktop but i think i could whip up a patch for that too.

Janleyx: what i mean to say for clarification, when you run steam on desktop mode the desktop configuration it has for the controller i believe its just gamepad, so no mouse on analog or anything like that is set by default to allow easy access and quality of life you know what i mean.

system_files/usr/bin/steamos-add-to-steam
Janleyx: Yeah i saw that it wasnt a part of the os and it makes sense since its essentially fedora. I use the steam deck as much as windows since during the day i use my pc and at night i use the steam deck so theres features on the steamos that are a must for this type of device and I feel this is a must.

Janleyx: Its a nice quality of life improvement and saves time.

README.md
Janleyx: yeah, I'm pretty new on the scene my bad. I love you're os by the way im always testing games on it and building plugins with it in mind. Theres one im working but i feel the os is too early in development for it. Id love to talk to you about it if you ever have some free time.

Janleyx:
What exactly do you need from me here? need me to divide theese into diferent pr's? the reason i have them all together its because i have them all under one patch.

JPyke3:
> What exactly do you need from me here? need me to divide theese into diferent pr's? the reason i have them all together its because i have them all under one patch.

Yes that's correct. From our perspective we need these features separate to allow for easier testing and review across a very broad set of devices.

Wouldn't have expected you to know this yet, as we have no contribution guide or pr template. We are hoping to fix this soon

Anything that really does belong to nested desktop can stay on this PR

Janleyx:
Ok ill work on it

Janleyx:
Apologies for the long delay here. I have been very busy lately, but I plan to revisit this patch soon, update it against the latest Armada release, and revalidate it on-device before requesting another review. Thank you for your patience.

PR #122, opened 2026-07-19
Add stick RGB lighting (HTR3212), reactive to sticks/buttons/screen color

Ga1dz1:
New Armada Control > Settings > Stick Lighting panel for devices with addressable RGB analog sticks (HTR3212 driver, 4 zones per stick). Confirmed live on Retroid Pocket 6 hardware; the LED zone layout there is identical to Retroid Pocket Mini V2's (unreleased device this feature was originally built for), despite the two being different SoC families (SM8550 vs SM8250) with different gamepad MCUs.

Ten modes: Static, Breathing, Battery (color follows charge level, solid green while charging), Battery + Breathing, Rainbow, Chase (a lit zone with a fading tail travels around each stick), Alternating (breathing, 180 degrees out of phase between sticks), Reactive (each stick's own deflection drives its brightness/hue, centered = off; each button press flashes both sticks in that button's own configurable color), Multidot (three colored dots chasing each other), and Ambilight (each stick tracks the average color of the screen near its own side, sampled off the DRM framebuffer via ffmpeg's kmsgrab). Most modes have per-mode Speed/ Intensity/Size sliders. A "follow screen brightness" toggle scales everything by the display's current backlight level.

Reactive and Ambilight need to read the gamepad/screen directly, alongside whatever InputPlumber and gamescope are already doing with them - not instead of. For the gamepad this means a second, non-exclusive open() of the same raw device node, which needs InputPlumber's composite device config to set passthrough: true on it (added to 02-ayn-controller.yaml, which also covers Retroid Pocket 6 alongside the AYN family) - this only works with InputPlumber >= the version carrying the fix in virtudude/armada-packages that makes passthrough apply to gamepad-classified source devices, not just keyboards (upstream's gamepad source implementation previously ignored the option entirely). A companion PR for that InputPlumber fix will follow. Retroid Pocket 6's raw gamepad device also happens to be mode 0000 (root-only) on real hardware, so leaving it ungrabbed doesn't risk Steam/SDL picking it up as a second controller.

armada-stick-led.service runs the whole thing continuously, gated by ConditionPathExists=/sys/class/leds/l:r1 so it's a no-op on hardware without this LED controller.

### DIFF
--- a/build_files/40-vendor-system-files.sh
+++ b/build_files/40-vendor-system-files.sh
@@ -29,6 +29,7 @@ systemctl enable armada-input-calibration.service
 systemctl enable armada-controller-type.service
 systemctl enable inputplumber.service
 systemctl enable armada-device-quirks.service
+systemctl enable armada-stick-led.service
 systemctl enable armada-fixups.service
 systemctl enable armada-installer-visibility.service
 systemctl enable armada-steamapps.service
--- a/decky/armada-control/main.py
+++ b/decky/armada-control/main.py
@@ -9,6 +9,13 @@ from armada_control.calibration import (
 )
 from armada_control.config import build_config
 from armada_control.controller import set_controller_type
+from armada_control.lighting import (
+    set_stick_led_color,
+    set_stick_led_flash_color,
+    set_stick_led_mode,
+    set_stick_led_param,
+    set_stick_led_screen_link,
+)
 from armada_control.power import save_power_config
 from armada_control.steam import installed_games
 from armada_control.system import set_ssh_enabled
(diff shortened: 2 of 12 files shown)


Ga1dz1:
This branch has drifted significantly from current `Ga1dz1/armada` main - the stick-lighting feature it introduces has since been reworked there:

- `battery`/`battery-breathing` modes -> replaced by an orthogonal `color_source` (static/battery) that combines with any mode, plus a togglable charging indicator
- `alternating` mode removed (lost its meaning once each stick is independently configurable)
- new `spin` mode (hard-edged single lit zone, no fading tail)
- gamma-corrected LED brightness output (the HTR3212 driver writes PWM linearly with no correction of its own)
- duotone color/position fixes confirmed against real hardware

Given how entangled this now is with unrelated later work in the same files (display menu, calibration), a clean minimal rebase needs hand-curation rather than a straight cherry-pick - flagging here so the "dirty" merge state has an explanation instead of silence. Will follow up with an updated branch.

Issue #31, opened 2026-06-24
Audio issues on resume


xXJSONDeruloXx:
Some games when resumed from fake-suspend seem to either have crackling audio, or more often it will sound like the audio is sped up slightly, or warbles between well paced and overly fast. 

Alsa/pulse lifecycle might need some scripting around different system events

virtudude:
> Some games when resumed from fake-suspend seem to either have crackling audio, or more often it will sound like the audio is sped up slightly, or warbles between well paced and overly fast. 
> 
> Alsa/pulse lifecycle might need some scripting around different system events

I've tried a few things to workaround this with limited success. The latest build has `pactl` installed if you want to experiment. I'm planning to revisit as well.

virtudude:
@xXJSONDeruloXx Do you have a game that repros this consistently? Of course now that I'm trying to fix it I can't get it to happen.

Roxoriku:
@virtudude I have the same issue. It seems to happen consistently with Into the Breach on my Odin 2 Mini.

Raxcoms:
Slightly different experience from what’s being reported.

I had my audio go completely silent on suspend/resume,
It came back after switching to desktop mode and then back to game mode.

I had no game running at the time. It was just idling on the SteamOS home page.

I use an Ayn Thor

awholedamnbear:
Adding to this, my issue is similar to @Raxcoms on my Odin 2 Portal. Resuming cuts out audio entirely, and in my case, it does not come back after switching between desktop and Game mode. It does come back on after restarting, but it's still lost any time I suspend. I am currently running the Preview OS update and the Stable channel Steam client.

justradical:
Is this still a problem or can this be closed?

PR #127, opened 2026-07-19
Fix audio corruption after fake-suspend resume, harden real-suspend wake

Ga1dz1:
fake-suspend's freeze_processes() used to freeze the whole user.slice, including pipewire/wireplumber themselves, for the sleep duration. But fake-suspend never touches the actual audio hardware, so the codec's clock keeps running the whole time - pipewire's own stream/clock bookkeeping went stale relative to the hardware ring buffer by the time it thawed, producing reported crackling/sped-up/warbling audio (armada issue #31), and in some cases outright silence in a stream that was already open before sleep (e.g. Steam's own UI sounds) while freshly-opened streams (e.g. a game launched after resume) were fine.

pipewire.service/wireplumber.service run in a sibling cgroup (session.slice) to gamescope/Steam's own cgroup, so freezing the whole user.slice was much broader than it needed to be. freeze_processes() now freezes only gamescope's own cgroup directly via its cgroup.freeze file, leaving pipewire/wireplumber running (silently, since audio is muted around the sleep window anyway) the entire time - nothing to resync on resume.

Also adds a systemd-sleep hook (50-armada-real-suspend) for devices using real suspend (ARMADA_SUSPEND_MODE != fake, currently AYN Odin 3): blanks stick RGB and mutes audio around the freeze/thaw edge, arms an optional RTC-backed watchdog so a device that can't wake on the power key still comes back on its own, and grabs the power key briefly after resume so the same physical press that wakes the kernel doesn't also read as a fresh short-press and put the device straight back to sleep. powerbuttond gets a backstop for the same rebound (REBOUND_GUARD_MS), in case the hook's grab loses the race with its own reader.

### DIFF
--- a/system_files/usr/lib/systemd/system-sleep/50-armada-real-suspend
+++ b/system_files/usr/lib/systemd/system-sleep/50-armada-real-suspend
@@ -1,0 +1,145 @@
+#!/bin/bash
+# systemd-sleep hook for real (non-fake) suspend, i.e. ARMADA_SUSPEND_MODE != fake
+# (currently AYN Odin 3, and SM8250 devices once s2idle wake is confirmed on
+# real hardware). Runs for every real suspend/resume, called by systemd-sleep
+# as "$0 pre|post suspend|...". Handles the things real suspend's kernel-level
+# device PM does NOT cover on its own:
+#
+#  - stick RGB: no kernel driver owns it, so it must be blanked/restored by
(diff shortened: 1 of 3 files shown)


virtudude:
Thanks for this PR. A few things have changed in fake-suspend but I pulled a couple of fixes that were still relevant here: https://github.com/virtudude/armada/commit/b581e8630404234a8355de6f1d4504e149d5a306

PR #156, opened 2026-07-25
External display support patch

EvilToasterDBU:
I commented out the forced screen output settings, and Gamescope started working perfectly with both the external display and internal screen.
If you wanna test it, check https://github.com/virtudude/armada/issues/29#issuecomment-5058788605, I’ve prepared a test build in my fork, along with a script to switch to the fork and back.

Tested on my Retroid Pocket 6.


justradical:
Hi, im already working on getting these vars out of there properly, while this does work on most devices it can still break when you dont define the outputs properly. I will close this but just know that it is being worked on

PR #161, opened 2026-07-26
Feat: add lsfg-vk decky

JPyke3:
**Note:** Depends on https://github.com/virtudude/armada-packages/pull/11 being merged first!

Adds Decky Lsfg-vk and adds it to the image. 

PR Notes:
* Bundles the official 0.12.5 release.
* Replaces the ARM build with Armada's own Package (Fixes issues with upstream compiling with GCC instead of Clang)
* Tested on AYN thor, everything validated working including frame gen on Outer Wilds

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,4 +1,5 @@
 ARG FEX_PKG=ghcr.io/virtudude/armada-packages/fex@sha256:5ac08826823c1228e5d25681c0508b79b01ff301e8c32d2cd5bbe72687615e7a
+ARG LSFG_VK_PKG=ghcr.io/virtudude/armada-packages/lsfg-vk@sha256:3912b6e8d5c98b973df85746ca58aa5c2188e33e3f8abc7e7618bfe284f4ad70
 ARG MESA_PKG=ghcr.io/virtudude/armada-packages/mesa@sha256:1b5bb7dec2d5cbcdbd3f895d3cdb6751c6a3a508b1fcc40e30cf83ad72229bcb
 ARG MANGOHUD_PKG=ghcr.io/virtudude/armada-packages/mangohud@sha256:583278eb4f9b55f6ee25bc1040ad8066d897e75b947ec8f2c73e01a360368e0c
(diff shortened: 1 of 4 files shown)


Containerfile
```diff
@@ -1,4 +1,5 @@
 ARG FEX_PKG=ghcr.io/virtudude/armada-packages/fex@sha256:5efff7dd05124e0653fd31a62bba78a68c87bd28f54ad12f6d0079acb3f07f7e
+ARG LSFG_VK_PKG=ghcr.io/jpyke3/armada-packages/lsfg-vk@sha256:dfd9a94d36e262fc0016753fae3ca40687efd1a71752caeae0b71ddcd01de425
```
JPyke3: Need to update this once the `armada-packages` merge is done :)

build_files/45-install-decky-plugins.sh
```diff
@@ -11,6 +11,26 @@ rm -f /usr/share/decky-plugins/armada-control/dist/*.map
...
+lsfg_sha256=13b8c8de5744a4fcf300e85971cb0c110f0734cb2db508c8de6309bbf8298a07
+lsfg_archive="$(mktemp)"
+lsfg_extract="$(mktemp -d)"
+
+curl --retry 3 --retry-delay 2 -fL -o "${lsfg_archive}" "${lsfg_url}"
+printf '%s  %s\n' "${lsfg_sha256}" "${lsfg_archive}" | sha256sum -c -
+unzip -q "${lsfg_archive}" -d "${lsfg_extract}"
+
+lsfg_src="${lsfg_extract}/Decky LSFG-VK"
+patch --directory="${lsfg_src}" --strip=1 \
```
virtudude, line 25: What will happen if someone updates the plugin via decky? They'd lose the patch?

JPyke3: Ah, good catch. `armada-decky-sync` runs on boot so in theory if a User did this you could reboot to revert.

But being prommpted every boot for updates would be annoying. I think I'll see if I can give it a unique vendor ID so that Decky won't treat this the same as upstream.

JPyke3:
*sigh* just changing the package name broke the plugin's functionality... Gotta look into this one a bit more after work today.

JPyke3:
Re-Tested on AYN thor after clean deletion of all `lsfg-vk` files and plugin. Reinstalls perfectly fine from a clean installation. I think this is good to merge

PR #166, opened 2026-07-26
Add Odin 3 stereo and virtual surround audio

enjihn:
## Summary

- Add device-gated Stereo and Virtual Surround Sound PipeWire profiles for the Odin 3 built-in speakers
- Keep the physical speaker sink internal, preserve the selected speaker profile across sessions, and hold the hardware route at unity gain
- Route Bluetooth and 3.5 mm outputs outside the speaker HRIR graphs, move active streams, and keep Steam's selected output synchronized
- Preserve the exact active internal output through real suspend, restore existing streams after resume, and resynchronize Steam without affecting an active external output
- Add managed installation and rollback, a checksum-pinned bundled HRIR, documentation, and focused tests

## Device scope

All setup and runtime helpers require an exact `ayn-odin-3`, `AYN Odin 3`, and `SM8750` policy match before changing audio state. Other Armada devices retain their existing audio configuration. The installed services perform only short device-policy checks and exit without changing the audio graph, routes, volume, or Steam selection when the gate does not match.

## Validation

- Focused policy, routing, Steam synchronization, and resume tests: 46 passed
- Linux resume-helper lifecycle tests passed
- Live Odin 3 audio checks passed with one Stereo sink, one Virtual Surround Sound sink, a 100 percent hidden hardware route, and Virtual Surround Sound restored as the selected output
- Physical deep suspend and resume passed on the Odin 3: the kernel completed deep suspend, the resume path ran exactly once, Virtual Surround Sound was restored, and Steam resynchronized to the current VSS device ID
- Post-resume eight-channel playback completed successfully through Virtual Surround Sound
- Audible eight-channel checks completed for both Stereo and Virtual Surround Sound
- Final tree matches the live-tested candidate and is based on current `main`

Physical Bluetooth and 3.5 mm connection tests remain pending, so this PR stays a draft.


JPyke3:
Seems draft has gone stale, closing

Issue #50, opened 2026-07-03
Quick Access Menu button does not work on AYANEO devices


ElberSousa:
I found an issue affecting multiple AYANEO devices.

The button that opens the SteamOS Quick Access Menu (the right-side menu button) does not work at all.

Tested devices

- AYANEO Pocket EVO
- AYANEO Pocket ACE
- AYANEO Pocket S2 Pro

The issue is present on all three devices.

I tested the same build on Retroid and AYN devices, where the button works correctly and opens the Quick Access Menu as expected.

This appears to be an issue specific to AYANEO devices.

Raxcoms:
That's a known issue listed on the readme
You can use home + a to open it for now

PR #169, opened 2026-07-28
Map Pocket EVO AYA SPACE, =, LC and RC buttons

jesherman:
These four buttons are wired to SoC GPIOs rather than to the controller MCU, so they never appear on the fake-Xbox-360 pad and were unmapped. They arrive instead on their own gpio-keys node, which the kernel only creates once the device tree declares it.

Kept out of the existing gpio-keys node deliberately: InputPlumber grabs the source devices it manages, so sharing that node would swallow volume up.

The device tree emits F13-F16 rather than AYANEO's stock KEY_A, KEY_EQUAL, KEY_LEFT and KEY_RIGHTSHIFT. Those type real characters, so any moment InputPlumber is not grabbing the node - desktop mode, a service restart - they would leak into whatever has focus.

Only AYA SPACE is bound to a fixed action, since QuickAccess is the one thing Steam Input cannot bind. The other three become paddles so they stay remappable per game.

Needs the matching aya-keys node in qcs8550-ayaneo-pocketevo.dts; until that lands nothing matches "aya-keys" and this config is inert.

Verified on hardware by patching the DTB in place and reading the emulated Steam Controller's HID reports: AYA SPACE opens QAM, and =, LC and RC report as L5, L4 and R4.

### DIFF
--- a/system_files/usr/share/inputplumber/capability_maps/ayaneo_pocket_evo_keys.yaml
+++ b/system_files/usr/share/inputplumber/capability_maps/ayaneo_pocket_evo_keys.yaml
@@ -1,0 +1,65 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+# Schema version number
+version: 2
+
+# The type of configuration schema
+kind: CapabilityMap
+
+# Name for the device event map
+name: AYANEO Pocket EVO GPIO Keys
+
+# Unique identifier of the capability mapping
+id: ayaneo_pocket_evo_keys
+
+# The Pocket EVO wires AYA SPACE, =, LC and RC directly to SoC GPIOs instead of
+# the controller MCU, so they arrive on their own gpio-keys node rather than on
+# the fake-Xbox-360 pad. The F13-F16 codes come from armada's aya-keys device
+# tree node; AYANEO's stock codes (KEY_A, KEY_EQUAL, KEY_LEFT, KEY_RIGHTSHIFT)
+# are deliberately not reused, since those type text if they ever reach a
+# compositor with InputPlumber stopped.
+#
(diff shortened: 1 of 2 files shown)


jesherman:
Kernel half is up: virtudude/armada-packages#14 — it adds the `aya-keys` gpio-keys node and its TLMM pinctrl state to `kernel/dts/qcs8550-ayaneo-pocketevo.dts`.

Merge order matters slightly: this PR is inert until that node exists, and that node is inert until `KERNEL_PKG` in the `Containerfile` is bumped to the rebuilt kernel image. So the sequence is armada-packages#14 → kernel image rebuild → `KERNEL_PKG` bump → this. Happy to send the digest bump once the image exists.

Both halves verified together on hardware: AYA SPACE opens QAM, and =, LC and RC report as L5, L4 and R4.

Macrey:
The Pocket S 2K needs the same treatment, if you're open to extending it.

Device tree:

    model:      AYANEO Pocket S 2K
    compatible: ayaneo,pocket-s1 qcom,sm8550

There is no aya-keys node, and the only gpio-keys node declares a single key:

    $ ls /proc/device-tree/gpio-keys/
    compatible  key-volume-up  name  pinctrl-0  pinctrl-names

The AYA button and the "two bars" button produce nothing at all, no event on any readable input node while pressing them, and nothing on the pad either: the controller source device declares only 11 buttons.

    $ EVIOCGBIT(EV_KEY) on the InputPlumber source node
    A, B, X, Y, L1, R1, Select, Start, Mode, L3, R3

So `01-ayaneo-controller.yaml` (matched here as "AYANEO Layout") applies `capability_map_id: ayaneo_mcu_xbox`, which waits on BTN5 for QuickAccess and BTN5 never arrives. Hence no QAM button on this device; Home + A is the
workaround, per the README and #50.

Build: beta 20260729.d812101, booted from microSD.

Happy to test a DTS node on hardware if that helps.

Issue #175, opened 2026-07-29
Improve fan curves and clocks


leonardomp10:
I'm on AYN Thor and the device heats a lot with official fan curves (80° in standby). I just changed them by myself and now the device is always under 70° even with load. Additionally I've improved a little the clocks of balanced profile. I'll put here the aggressive fan curves and balanced clocks I'm using. Battery life was not affected at all.

[profile.balanced]
label=Balanced
cpu_governor=schedutil
cpu_max=0.85
cpu_underclock=medium
gpu_max=1.0
gpu_min=0.5
fan_curve=aggressive

[fan_curve.aggressive]
label=Aggressive
curve=90:255,85:255,80:255,70:255,65:150,60:100,50:75,30:51,0:51

[underclock.SM8550.medium]
cpu_max_policy0=1655200
cpu_max_policy3=2054400
cpu_max_policy7=2192800

raptor-chicken:
this is a good idea!

PR #177, opened 2026-07-30
fix(fan_curves): reduce heat SM8550 issue #175

leonardomp10:
This fix improves aggressive fan curves to be really aggressive. Additionally changes balanced profile definitions and increases a little SM8550 clocks under medium underclock. Tested on AYN Thor and reduced around 10 degrees even under load (max 70-75 degrees). Battery life was not affected by the changes.


JPyke3:
Closing as #260 handles fan control

PR #179, opened 2026-07-30
feat: add CPU turbo boost toggle

pacoa-kdbg:
## Summary
- Add CPU boost/turbo controls to `armada-powerd` and expose them via D-Bus status/properties
- Add `armada-power boost [on|off]` CLI support
- Surface a Decky Power tab toggle that disables itself when the device has no CPU boost sysfs control

## Test Plan
- `python3 -m py_compile system_files/usr/libexec/armada/armada-powerd system_files/usr/bin/armada-power decky/armada-control/main.py decky/armada-control/py_modules/armada_control/*.py`
- `cd decky/armada-control && npm install && npm run build`
- Direct shell equivalent of `just test` because `just` is not installed on this VM:
  ```bash
  shopt -s nullglob
  for test_file in tests/*-test.sh; do
    bash "$test_file"
  done
  ```

### DIFF
--- a/decky/armada-control/main.py
+++ b/decky/armada-control/main.py
@@ -9,7 +9,7 @@ from armada_control.calibration import (
 )
 from armada_control.config import build_config
 from armada_control.controller import set_controller_type
-from armada_control.power import save_power_config
+from armada_control.power import save_power_config, set_cpu_boost_enabled
 from armada_control.steam import installed_games
 from armada_control.system import set_ssh_enabled
 from armada_control.tweaks import load_compat_applied, save_compat_applied, save_tweaks
@@ -25,6 +25,10 @@ class Plugin:
 
     async def save_power_config(self, data):
         await asyncio.to_thread(save_power_config, data)
+        return await self.get_config()
+
+    async def set_cpu_boost_enabled(self, enabled):
+        await asyncio.to_thread(set_cpu_boost_enabled, enabled)
         return await self.get_config()
 
     async def save_tweaks(self, data):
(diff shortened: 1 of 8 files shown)


pacoa-kdbg:
Pocket DS live validation (Armada 20260728.c859b5b / kernel 7.0.11):

- Device exposes CPU boost controls:
  - `/sys/devices/system/cpu/cpufreq/boost`
  - `/sys/devices/system/cpu/cpufreq/policy0/boost`
  - `/sys/devices/system/cpu/cpufreq/policy3/boost`
  - `/sys/devices/system/cpu/cpufreq/policy7/boost`
- Ran patched `armada-powerd` via a temporary systemd override and validated:
  - `armada-power status` reports `cpu_boost_available=True`
  - `armada-power boost on` sets global boost to `1` and reports `cpu_boost_enabled=True`
  - daemon restart preserves/reapplies selected boost state
  - `armada-power boost off` restores global/per-policy sysfs values to `0`
- Ran the patched Decky backend against the live daemon:
  - initial config: `{'available': True, 'enabled': False}`
  - `set_cpu_boost_enabled(True)` returned enabled state and sysfs changed
  - `set_cpu_boost_enabled(False)` returned disabled state and sysfs restored to off
- Restarted `plugin_loader.service`; Armada Control loaded successfully and the built frontend bundle contains the new `CPU Turbo Boost` row.

One hardware-specific issue found and fixed during validation: Pocket DS accepts writes to the global boost node, but some per-policy boost nodes return `EINVAL`; the patch now prefers the global boost node when present and only falls back to per-policy nodes when no global node exists.

PR #184, opened 2026-07-31
Docs: Remove vendor specific callout in README

### LINKED ISSUE #157: Remove vendor specific callout in README
Balena etcher is not good to use or recommend to anyone and hasn't been for a long time. https://web.archive.org/web/20250829230423/https://tails.net/news/rufus/index.en.html Instead readme should suggest Rufus for windows, or Pi Imager or the many other easy to use options on Linux.

JPyke3:
Closes #157

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ internal storage so it boots without the card (see
 
 1. Flash the Armada image to SD.
 
-   Use Balena Etcher to flash the latest `armada-YYYYMMDD.img.gz` image to a
+   Use a disk imaging tool to write the latest `armada-YYYYMMDD.img.gz` image to a
    64GB or larger SD card (A2 speed for best results).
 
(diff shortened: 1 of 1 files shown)


virtudude:
What about listing options instead? ROCKNIX has: 

`Common imaging tools include [Balena Etcher](https://www.balena.io/etcher/), [Raspberry Pi Imager](https://www.raspberrypi.com/software/), and [Win32 Disk Imager](https://sourceforge.net/projects/win32diskimager/). If you're skilled with the command line, dd works fine too.`

JPyke3:
Happy to do that! Just had the draft in place to address the issue.

Ill probably do exactly this 👍

JPyke3:
I think we are going to catch this fix in the switch to the wiki - So we will update it accordingly there. Once the wiki is going then we can just link off to it in README. I don't think there is any action that needs to be done here

PR #185, opened 2026-08-01
Prototype touchscreen trackpad mode

gh123man:
## What changed

- add a fail-safe system service that grabs the primary touchscreen only when trackpad mode is enabled and emits a standard relative virtual mouse
- support one-finger pointer movement, single-tap left click, two-finger-tap right click, and double-tap-and-hold dragging
- rotate raw touch coordinates using Armada's per-device panel orientation, including the portrait-native Retroid panels
- add a persisted `Touchscreen Trackpad` toggle under Armada Control's Advanced > Input section
- add focused gesture, orientation, syntax, package, and service-installation tests

## Why

Steam Input does not expose a handheld's physical touchscreen as a configurable trackpad, and InputPlumber's touchscreen-to-mouse translation is not implemented in the pinned package. A small Armada-owned prototype lets us validate the interaction model without first coordinating an InputPlumber package change.

Direct touch remains the default. In trackpad mode the service uses an exclusive evdev grab, so Gamescope does not also receive absolute touches. If the service exits or crashes, the kernel releases the grab and direct touch becomes available again.

## User impact

Users can switch modes from the controller-accessible Decky quick-access menu. Trackpad mode works as a conventional system mouse in Steam, games, and Desktop Mode rather than relying on a per-game Steam Input layout.

## Validation

- all repository shell tests passed in a disposable Fedora 44 ARM64 container
- six focused gesture and panel-orientation unit tests passed
- Decky production bundle built successfully with the locked dependencies
- Python and shell syntax checks passed
- `git diff --check` passed
- Fedora 44's official repositories provide `python3-evdev`

A full Armada image was not built locally. The pull-request container build should provide that integration check.

## Hardware validation still needed

- verify pointer direction and speed on a left-rotated Retroid and a right-rotated handheld
- verify direct/trackpad switching in Gaming Mode and Desktop Mode
- verify tap, two-finger tap, double-click, and double-tap drag behavior
- verify stopping the service immediately restores direct touch
- check interaction with Steam overlays, suspend/resume, and device reconnects

This is intentionally a prototype. Sensitivity, scrolling, and per-game overrides can be added after the basic interaction is tested on hardware.

### DIFF
--- a/build_files/10-base-packages.sh
+++ b/build_files/10-base-packages.sh
@@ -24,6 +24,7 @@ dnf5 -y install --setopt=install_weak_deps=False \
     bluez \
     dbus-broker \
     python3-gobject \
+    python3-evdev \
     python3-websocket-client \
     polkit \
     upower \
--- a/build_files/40-vendor-system-files.sh
+++ b/build_files/40-vendor-system-files.sh
@@ -35,6 +35,7 @@ systemctl enable sddm.service
 systemctl enable armada-session-default.service
 systemctl enable seatd.service
 systemctl enable armada-input-calibration.service
+systemctl enable armada-touchscreen-trackpad.service
 systemctl enable armada-controller-type.service
 systemctl enable inputplumber.service
 systemctl enable armada-device-quirks.service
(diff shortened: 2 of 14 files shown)


JPyke3:
This draft has been here a while, closing as stale

PR #187, opened 2026-08-01
Speed up DNF during image builds

tslmy:
## Summary

This adds an early build step that configures libdnf5 for faster and more resilient package downloads during image builds:

- enables `fastestmirror`
- raises `max_parallel_downloads` to 10
- sets a minimum transfer rate and timeout so very slow mirrors fail over sooner
- adds an optional `ARMADA_FEDORA_MIRROR` build arg for builders who know a nearby Fedora mirror

By default this does not hard-code a regional mirror. The explicit mirror override is only applied when `ARMADA_FEDORA_MIRROR` is set, and it only rewrites Fedora repo files that use Fedora standard commented `baseurl` template.

The build now writes the libdnf5 speed-up config at the start of the container build, optionally rewrites Fedora repo files when `ARMADA_FEDORA_MIRROR` is set, and then runs a late restore step after the final DNF operations. That restore step removes `/etc/dnf/libdnf5.conf.d/10-armada-build.conf` and restores any Fedora repo files it temporarily changed before the image is finalized.

So the installed OS should keep Fedora's normal repository configuration, while local/CI image builds can still use faster parallel downloads and an operator-selected nearby mirror.

## Speed-up observed

Tested from the western US while building in a Fedora Lima VM on macOS. With an explicit nearby mirror, DNF download phases that had previously selected slow mirrors or stalled in the hundreds of KiB/s range were observed around 6-8 MiB/s on representative transactions. One later DNF transaction in the successful image build downloaded 43 MiB at about 7.8 MiB/s.

The total build-time improvement depends heavily on cache state and non-DNF work, especially package/image assembly, but the DNF download portions were roughly an order of magnitude faster in this test environment.

## Usage

The general build path gets the libdnf5 tuning automatically. Builders who want to pin a known-good mirror can run, for example:

```bash
ARMADA_FEDORA_MIRROR=https://mirrors.kernel.org/fedora just build localhost/armada test
```

Builders elsewhere can use a geographically appropriate Fedora mirror, or leave `ARMADA_FEDORA_MIRROR` unset and rely on Fedora mirror selection with `fastestmirror` enabled.

## Validation

- `bash -n build_files/05-configure-dnf.sh build_files/build.sh`
- `just --fmt --check`
- completed a local image build in a Fedora Lima VM from the western US using `ARMADA_FEDORA_MIRROR=https://mirrors.kernel.org/fedora`

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -43,6 +43,7 @@ COPY system_files /system_files/
 
 FROM quay.io/fedora/fedora-bootc:44
 ARG ARMADA_VERSION=unknown
+ARG ARMADA_FEDORA_MIRROR=
 LABEL org.opencontainers.image.version="${ARMADA_VERSION}"
 
 RUN --mount=type=bind,from=ctx,source=/,target=/ctx \
@@ -66,6 +67,7 @@ RUN --mount=type=bind,from=ctx,source=/,target=/ctx \
     --mount=type=tmpfs,dst=/tmp \
     mkdir -p /usr/lib/armada && \
     printf '%s\n' "${ARMADA_VERSION}" >/usr/lib/armada/version && \
+    export ARMADA_FEDORA_MIRROR="${ARMADA_FEDORA_MIRROR}" && \
     /ctx/build_files/build.sh
 
 RUN bootc container lint
--- a/Justfile
+++ b/Justfile
@@ -82,6 +82,9 @@ build $target_image=image_name $tag=default_tag:
     BUILD_ARGS=()
     ARMADA_VERSION="$(TZ=America/New_York date +%Y%m%d).$(git rev-parse --short HEAD)"
     BUILD_ARGS+=("--build-arg" "ARMADA_VERSION=${ARMADA_VERSION}")
+    if [[ -n "${ARMADA_FEDORA_MIRROR:-}" ]]; then
+        BUILD_ARGS+=("--build-arg" "ARMADA_FEDORA_MIRROR=${ARMADA_FEDORA_MIRROR}")
+    fi
 
     # Allow local armada-packages images to override pinned package images.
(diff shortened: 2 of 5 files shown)


JPyke3:
Hey Mate, appreciate the PR. You mention specifically that you're optimizing for buildtime. However, it seems this PR is going to ship these dnf changes to users?

I've always found dnf to have pretty sensible defaults and I live in Au where out internet is not renowned for speed. 

So as is I'd be a bit reluctant to merge as I don't think built time needs to be sped up at the expense of user defaults.

Happy to hear your thoughts though!

tslmy:
Hey @JPyke3 , thanks for responding this quickly. 

Totally agree that we should avoid shipping user defaults. As you see, I made a point of making `ARMADA_FEDORA_MIRROR` a variable instead of hardcoding it. As for `fastestmirror=True` -- I _had thought_ that it was harmless to ship as a default.

But now that you asked, I got curious why `fastestmirror` defaults to `False`. [Turns out](https://discussion.fedoraproject.org/t/fastest-mirror/135493/5) it picks by TCP socket latency (not necessarily the fastest) and can override mirror-manager’s own ordering, including bandwidth/preferred-mirror hints. Indeed, that's not I'd like to ship to end users.

> pretty sensible defaults

Right, TIL about "Fedora MirrorManager" (bear with me please -- I'm on a Mac and have little experience with DNF 😅 ). Somehow, in my western-US build environment, it appears to have selected one (or more) mirrors with poor sustained throughput. (https://terrapkg.com/ 500'd on me once.) Pinning at `mirrors.kernel.org` avoided that path and gave ~6-8 MiB/s.

Back to this PR. I've added a script to remove that DNF conf file and `trap` it on `EXIT` in `build.sh`.

This makes the override ephemeral, which feels less worthy of a PR merge than I'd previously thought. So totally cool if we don't end up merging it (I can still keep this change local; I'm not missing out on anything). Thanks!

JPyke3:
@tslmy appreciate the contribution on this one but will close as per the discussion above :)

Issue #22, opened 2026-06-21
Controller support on desktop mode


Soraasinsky:
First of all, I wanna thank you for your amazing work. This is working very well despite being a new project. Currently testing on my KONKR Pocket Fit (8gen3) and I can confirm that it is working pretty good. Actually, there's been one bug where, whenever I start Armada, I need to enter and leave suspend mode (clicking on the power button twice)  so the screen turns on.

But that's not what I've come to talk. Since ROCKNIX device support currently doesn't implement touchscreen controls on the KONKR, I don't know how to navigate into desktop mode, and I guess we currently can't.

May I ask you, is it possible to implement controller navigation, just like in the Steam Deck? Where the left joystick becomes the mouse control and R2 is the right mouse button. That and Steam's virtual keyboard would make the desktop mode perfectly usable.

Roxoriku:
I don't know for certain but I assume that if we can get Steam as a startup application the desktop controller profile should activate and function as you describe.

In the meantime, do you have access to a usb-c keyboard/mouse or a usb-c to usb-a adapter + a usb-a KB/M?
That can work as a temporary solution.

virtudude:
I've been working on this and I have a lot of the pieces in place but haven't gotten the setup exactly right. The main issue is getting the Steam keyboard to behave as seamlessly as the KDE onscreen keyboard, which just pops up on text input. Also convincing Steam that the controller type is supported but I think I have a fix there. I'm releasing an update today that's focused on game mode and will be pivoting back to polishing desktop mode very soon.

Soraasinsky:
@Roxoriku 
> In the meantime, do you have access to a usb-c keyboard/mouse or a usb-c to usb-a adapter + a usb-a KB/M?In the meantime, do you have access to a usb-c keyboard/mouse or a usb-c to usb-a adapter + a usb-a KB/M?

I do own a mouse and an usb-c to usb-a adapter and it's working seamlessly, but I don't own an keyboard just yet (I use my built-in keyboard on my laptop) and I can't get KDE's virtual keyboard to work for some reason.

@virtudude 
> I've been working on this and I have a lot of the pieces in place but haven't gotten the setup exactly right. The main issue is getting the Steam keyboard to behave as seamlessly as the KDE onscreen keyboard, which just pops up on text input. Also convincing Steam that the controller type is supported but I think I have a fix there. I'm releasing an update today that's focused on game mode and will be pivoting back to polishing desktop mode very soon.

[...]

PR #160, opened 2026-07-26
fix: keep Pocket DS dual screen desktop-only

pacoa-kdbg:
## Summary
- Add AYANEO Pocket DS dual-screen metadata in the same style as AYN Thor: primary/top panel on `DSI-1`, secondary/bottom panel on `DSI-2`, touchscreen names, and shared `ARMADA_PANEL_ORIENTATION=left`.
- Keep Game Mode primary-panel-only by inhibiting the lower Goodix touchscreen in the Steam/gamescope session and explicitly disabling the secondary KScreen output during desktop → Game Mode handoff.
- Keep Desktop Mode dual-screen by making `desktop-bootstrap` re-run `setup-dual-screen` for dual-screen devices and by having `setup-dual-screen` enable and rotate both outputs before positioning them.
- Make missing touchscreen devices non-fatal so display layout still succeeds if the lower Goodix device probes late or is absent.

## Companion PR
- Kernel/DTS/backlight side: https://github.com/virtudude/armada-packages/pull/10
- This userspace PR expects the companion kernel/package PR so the Pocket DS lower DRM connector/backlight path exists and powers down correctly.

## Test Plan
- `python3 -m py_compile system_files/usr/libexec/armada/setup-dual-screen`
- `bash -n system_files/usr/libexec/armada/session-control system_files/usr/libexec/armada/desktop-bootstrap system_files/etc/gamescope-session-plus/sessions.d/steam system_files/usr/libexec/armada/device-env`
- `git diff --check`
- `for test_file in tests/*-test.sh; do bash "$test_file"; done`
  - `Pocket DS desktop-only dual-screen checks passed`
  - `powerbutton resume and cleanup test passed`
  - `missing touchscreen is non-fatal`

## Live Pocket DS validation
Validated with the companion kernel hotfix on real AYANEO Pocket DS hardware:
- Game Mode:
  - `DSI-2`: disabled / DPMS Off
  - `sy7758-backlight`: `bl_power=4`
  - lower Goodix touch: inhibited
  - physical result: lower screen/backlight dark
- Desktop Mode:
  - `DSI-1`: enabled / On
  - `DSI-2`: enabled / On
  - `sy7758-backlight`: `bl_power=0`
  - lower Goodix touch: uninhibited
  - physical result: bottom screen lit and correctly oriented

## Notes
- PR CI validates the image build. Publishing/signing/installability still follows the repository workflow policy for PR events.
- Temporary broad sudo used during live validation was removed after testing.

PR #194, opened 2026-08-02
[For dual-display devices] Pin virtual keyboard to the bottom screen

### LINKED ISSUE #206: Add the ability for users to pin their virtual keyboard on the second screen for dual screen devices
In Android, users have a way to optionally set their virtual keyboard to always appear on the secondary screen so the primary screen is always free for viewing.

Armada could benefit of a similar functionality

tslmy:
## Summary

This configures Armada to consume a patched KWin package and use per-device configuration to pin Plasma's virtual keyboard to a specific display connector.

On dual-screen handhelds such as the AYN Thor, the intended behavior is for the on-screen keyboard to appear on the bottom display instead of following KWin's active-output heuristic. The Thor device config sets:

```sh
ARMADA_VIRTUAL_KEYBOARD_CONNECTOR=DSI-1
```

The Plasma session exports that device setting before starting KWin. Devices that leave the setting empty keep upstream KWin's current active-output behavior.

Tested with an AYN Thor:

| Before | After |
| --- | --- |
|  Screenshot 2026-08-02 at 16 41 49 | Screenshot 2026-08-02 at 16 41 57 |

## Package Dependency

Per review feedback, the KWin patch has been moved to `armada-packages`: https://github.com/armada-os/armada-packages/pull/22 . This PR now consumes the KWin package image instead of rebuilding KWin inside the Armada base image. 

> [!WARNING]
> The current `KWIN_PKG` value uses `:latest` until the package PR publishes a digest that can be pinned here.

## Testing

```sh
bash -n build_files/build.sh build_files/10-base-packages.sh system_files/usr/libexec/armada/device-env system_files/usr/libexec/armada/start-plasma tests/thor-virtual-keyboard-test.sh
tests/thor-virtual-keyboard-test.sh
```


pacoa-kdbg:
Pocket DS follow-up: #195 configures `ARMADA_VIRTUAL_KEYBOARD_CONNECTOR=DSI-2`, using the verified lower-display connector from #160. It requires this PR’s patched-KWin build path.

JPyke3 (changes requested):
Hey all,

I do like the changes here. But I believe we should package kwin just like we package the other components in [armada-packages](https://github.com/armada-os/armada-packages) instead of in the base repo.

Also - I feel we shouldn't be using built-time flags to manage things like if the image should ship with components specific to the thor. Ideally we would ship one image that everyone can consume and use per-device configs to manage the differences.

The expected structure would therefore be:

* an `armada-packages` PR that builds and publishes the patched KWin package;
* this `armada` PR updated to consume that package;
* the per-device `ARMADA_VIRTUAL_KEYBOARD_CONNECTOR` configuration retained here.

I’m requesting changes rather than closing this, since the feature itself looks useful and the existing PR can be reworked around that packaging model.

tslmy:
Great idea @JPyke3 ! I wasn't aware of the existence of https://github.com/armada-os/armada-packages, and I love the idea of it. It honestly also makes my life easier if and when I choose to upstream this feature to [KWin itself](https://bugs.kde.org/show_bug.cgi?id=523755).

PR updated. Tested on real device with an image built from [fe6b0ee](https://github.com/armada-os/armada/commit/fe6b0ee0eb5dbafbe7bf2ca0d75ddd917bcf170e).

JPyke3 (approved):
Looks all good now! Thanks for splitting it up! Unfortunately I won't have time to merge this one in the next few hours (waiting for kwin to compile) but when I wake up I'll get on it tomorrow :D

tslmy:
Thank you @JPyke3 ! This is my first merged PR to this repo. I'm elated.

PR #195, opened 2026-08-03
feat: pin Pocket DS virtual keyboard to lower display

pacoa-kdbg:
## Summary
- Configure the AYANEO Pocket DS virtual keyboard output as its lower physical display: `DSI-2`.
- Add a regression test for this device mapping.

## Related pull requests
- Depends on [#160](https://github.com/armada-os/armada/pull/160) for the Pocket DS dual-screen desktop policy and its established `DSI-1` (top) / `DSI-2` (bottom) mapping.
- Requires [#194](https://github.com/armada-os/armada/pull/194), which adds the optional patched-KWin implementation that reads `ARMADA_VIRTUAL_KEYBOARD_CONNECTOR`.

After #160 and #194 are included in an image built with #194’s `build_patched_kwin` option, Plasma will pin Pocket DS’s virtual keyboard to the bottom `DSI-2` display instead of following the active-output heuristic.

## Test plan
- `bash tests/pocket-ds-virtual-keyboard-test.sh`
- `for test_file in tests/*-test.sh; do bash "$test_file"; done`
- `git diff --check`

### DIFF
--- a/system_files/usr/lib/armada/devices/ayaneo-pocket-ds.conf
+++ b/system_files/usr/lib/armada/devices/ayaneo-pocket-ds.conf
@@ -1,7 +1,11 @@
 ARMADA_DEVICE_ID=ayaneo-pocket-ds
 ARMADA_DEVICE_NAME='AYANEO Pocket DS'
-ARMADA_SOC_CLASS=SM8550
 
+# Plasma's virtual keyboard belongs on the lower Pocket DS display. Consumed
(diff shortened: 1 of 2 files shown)


JPyke3:
Thanks for the PR. 

Marking as blocked because we are waiting on some architectural changes in #194.

justradical:
im pretty sure this can be closed, the screen should be pinning on the pocket ds

Issue #19, opened 2026-06-18
File Picker not opening in Steam on Desktop Mode


Raxcoms:
When I try to add a non steam app in desktop mode and click “Explore” to pick a file, Dolphin doesn’t launch as expected.

This issue prevents me from browsing and selecting a custom file.

virtudude:
Thanks for the report. I think I see the issue. Will get it fixed soon.

virtudude:
I'm still looking into this. Turned out to be trickier than I expected. It's possible it's a Steam ARM bug but I'll keep at it.

Raxcoms:
Thanks for the update 🙂

PaulWhitfieldJr:
Use this decky plugin as a workaround
[](https://github.com/javierfurus/naiseika) It's mainly for rocknix but it does work on Armada OS

PaulWhitfieldJr:
https://github.com/javierfurus/naiseika

Raxcoms:
So I noticed Firefox also can't open Dolphin when trying to show a file in folder

Trying to google about it I found something about DBUS being misconfigured but I couldn't verify it

virtudude:
> So I noticed Firefox also can't open Dolphin when trying to show a file in folder
> 
> Trying to google about it I found something about DBUS being misconfigured but I couldn't verify it

I tried to repro this with Firefox in KDE but could not. Can you go to https://gofile.io in Firefox and click the `Drag & Drop or Click to Upload` button and let me know what happens? It opens the file picker for me.

Raxcoms:
> I tried to repro this with Firefox in KDE but could not. Can you go to https://gofile.io in Firefox and click the `Drag & Drop or Click to Upload` button and let me know what happens? It opens the file picker for me.

The file picker from gofile did work
The picker when select a download's location is also working

The "show in folder" button from downloaded files that should open dolphin in the file's storage location isn't working, dolphin doesn't open 

It though it might be related, but it might be different since the picker itself works

CorruptMilkshake:
looking for a workaround i found a post suggesting changing the default file manager with the inode/directory line in mimeapps.list. of course i was unable to test this, but i did see it points to nautilus, not dolphin as i assume it ought to.

keks2293:
> looking for a workaround i found a post suggesting changing the default file manager with the inode/directory line in mimeapps.list. of course i was unable to test this, but i did see it points to nautilus, not dolphin as i assume it ought to.

[...]

PR #209, opened 2026-08-03
move splash files to armada-packages

justradical:
depends on https://github.com/armada-os/armada-packages/pull/19


justradical:
build here will fail until the correct package is under it

PR #214, opened 2026-08-04
Start the ADSP when its boot-time firmware load loses the race

jesherman:
## Problem

`remoteproc` requests the ADSP firmware at **~3s into boot, inside the initramfs** — about 4s before `/boot` mounts:

```
[    3.197727] remoteproc remoteproc0: Direct firmware load for qcom/sm8550/ayaneo/adsp.mdt failed with error -2
[    3.199251] remoteproc remoteproc0: request_firmware failed: -2
```

dracut ships `qcom_q6v5_pas.ko` in the initramfs but **no qcom firmware** (`lsinitrd | grep -c lib/firmware/qcom` → 0), so the probe fails and remoteproc never retries after switch-root. The ADSP hosts the battery manager behind `pmic_glink`, so every `power_supply` attribute then reads back empty and Steam shows 0% on a charged device.

Not a race — it fails on every boot I have logs for. What varies is whether anything starts the ADSP afterwards. Full timing data across four boots is in the comment below.

## Fix

Oneshot that writes `start` once the filesystem is up, ordered before `armada-powerd` and the session. No-op if already running, 3 retries, never fails the boot.

## Testing

Pocket EVO, reproduced on a losing boot and recovered first try:
```
adsp-recover: adsp is offline; starting it
adsp-recover: adsp started (attempt 1)
→ cap=78 status=Discharging
```

## Alternative worth considering

Fixing it in dracut instead — omit `qcom_q6v5_pas` from the initramfs so it loads from the rootfs where the firmware exists, or include `/usr/lib/firmware/qcom`. That prevents the failed probe rather than repairing it. This PR is the more robust option (recovers regardless of cause); happy to do either or both.

### DIFF
--- a/build_files/40-vendor-system-files.sh
+++ b/build_files/40-vendor-system-files.sh
@@ -29,6 +29,7 @@ systemctl enable armada-input-calibration.service
 systemctl enable armada-controller-type.service
 systemctl enable inputplumber.service
 systemctl enable armada-device-quirks.service
+systemctl enable armada-adsp-recover.service
 systemctl enable armada-fixups.service
 systemctl enable armada-installer-visibility.service
 systemctl enable armada-steamapps.service
--- a/system_files/usr/lib/systemd/system/armada-adsp-recover.service
+++ b/system_files/usr/lib/systemd/system/armada-adsp-recover.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=armada ADSP recovery
+# remoteproc requests the ADSP firmware early and does not retry on failure, so
+# run once the rootfs is definitely mounted. Ordered before the battery and
+# session consumers so they do not read an empty power supply.
+After=sysinit.target local-fs.target
+Before=armada-powerd.service display-manager.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/libexec/armada/adsp-recover
+StandardOutput=journal
+StandardError=journal
+
+[Install]
+WantedBy=multi-user.target
--- a/system_files/usr/libexec/armada/adsp-recover
+++ b/system_files/usr/libexec/armada/adsp-recover
@@ -1,0 +1,40 @@
+#!/bin/bash
+# The ADSP hosts the battery manager firmware behind pmic_glink. Its firmware
+# request can fire before the rootfs holding /usr/lib/firmware is reachable;
+# remoteproc logs "request_firmware failed: -2" and never retries, so the ADSP
+# stays offline for the whole boot. Every power_supply attribute then reads back
+# empty and the battery looks absent (Steam shows 0%). Losing the race is
+# intermittent, so start it here once the filesystem is definitely up.
+set -uo pipefail
+
+log() { echo "adsp-recover: $*"; }
+
+found=0
+for rp in /sys/class/remoteproc/remoteproc*; do
+    [[ -r "${rp}/name" && -w "${rp}/state" ]] || continue
+    [[ "$(cat "${rp}/name" 2>/dev/null)" == "adsp" ]] || continue
+    found=1
+
+    state="$(cat "${rp}/state" 2>/dev/null)" || continue
+    if [[ "$state" == "running" ]]; then
+        log "adsp already running"
+        exit 0
+    fi
+
+    log "adsp is ${state}; starting it"
+    for attempt in 1 2 3; do
+        echo start > "${rp}/state" 2>/dev/null || true
+        sleep 2
+        if [[ "$(cat "${rp}/state" 2>/dev/null)" == "running" ]]; then
+            log "adsp started (attempt ${attempt})"
+            exit 0
+        fi
+    done
+
+    # Non-fatal: a device without working audio/battery still boots to Steam.
+    log "failed to start adsp after 3 attempts"
+    exit 0
+done
+
+(( found )) || log "no adsp remoteproc present; nothing to do"
+exit 0


jesherman:
Reproduced and validated again on an AYANEO Pocket EVO, on a fresh boot that lost the race:

```
$ cat /sys/class/remoteproc/remoteproc0/state
offline
$ cat /sys/class/power_supply/battery/capacity
cat: .../capacity: Resource temporarily unavailable
```

Note the read returns `EAGAIN` rather than an empty file in this instance — either way, consumers see no battery.

Running the script from this PR against that state:

```
adsp-recover: adsp is offline; starting it
adsp-recover: adsp started (attempt 1)
adsp=running   cap=78   status=Discharging
```

Recovered on the first attempt, and a second run correctly reports `adsp already running` and exits 0, confirming the no-op path.

virtudude:
@jesherman can you reshare the firmware load failure logs with timestamps? I'm curious to know more about how far into boot this is happening

jesherman:
i should note this issue specifically arose when i booted after a full drain to 0

Logs with monotonic timestamps, and the mechanism turns out to be more specific than "a race" — I need to correct my earlier description.

### Timing

The failure lands at **~2.9–3.2 s into boot**, consistently. Four consecutive boots on a Pocket EVO:

```
boot -1  [    3.196435] remoteproc remoteproc0: adsp is available
         [    3.197727] remoteproc remoteproc0: Direct firmware load for qcom/sm8550/ayaneo/adsp.mdt failed with error -2
         [    3.198071] remoteproc remoteproc0: powering up adsp
         [    3.198795] remoteproc remoteproc0: Direct firmware load for qcom/sm8550/ayaneo/adsp.mdt failed with error -2
         [    3.199251] remoteproc remoteproc0: request_firmware failed: -2

boot -2  [    3.040492] adsp is available
         [    3.040642] ... failed with error -2
         [    3.042474] request_firmware failed: -2

boot -3  [    3.170453] adsp is available
         [    3.175688] request_firmware failed: -2

boot -4  [    2.867544] adsp is available
         [    2.868194] request_firmware failed: -2
```

For context on where that sits in boot:

```
[    0.93–1.19] Freeing unused kernel memory        <- end of kernel init, initramfs starts
[    2.87–3.20] adsp firmware load fails (-2)       <- HERE
[    7.33–7.50] EXT4-fs (sda13): mounted /boot
[    9.63–12.9] EXT4-fs (mmcblk0p1): mounted
```

So it fails **inside the initramfs, ~4 s before /boot is even mounted** and before the real rootfs is available.

### Mechanism

It's not timing-dependent — it's deterministic given what's in the initramfs:

```
$ lsinitrd initramfs-7.0.11.img | grep -E 'q6v5|battmgr'
usr/lib/modules/7.0.11/kernel/drivers/remoteproc/qcom_q6v5_pas.ko
usr/lib/modules/7.0.11/kernel/drivers/remoteproc/qcom_q6v5.ko
usr/lib/modules/7.0.11/kernel/drivers/power/supply/qcom_battmgr.ko

$ lsinitrd initramfs-7.0.11.img | grep -c lib/firmware/qcom
0
```

dracut ships `qcom_q6v5_pas.ko` in the initramfs but **no qcom firmware at all**. udev loads the driver, it probes at ~3 s, `request_firmware()` cannot find `qcom/sm8550/ayaneo/adsp.mdt`, and remoteproc's auto-boot does not retry after switch-root. The ADSP then stays `offline` for the whole boot.

Correction to my PR description: I called this intermittent. The *initial load* fails on every boot I have logs for. What actually varies is whether anything starts the ADSP afterwards — the later `powering up adsp` entries in these boots (at 76 s, 160 s, 234 s, 354 s) are all me starting it by hand over SSH, not spontaneous recovery.

### Alternative fix worth considering

Given the mechanism, there's a case for fixing it in dracut instead of at runtime — either omitting `qcom_q6v5_pas` from the initramfs so it loads from the rootfs after switch-root where the firmware exists, or including `/usr/lib/firmware/qcom` in the initramfs. That would avoid the failed probe entirely rather than repairing it after the fact.

The oneshot in this PR is still the more robust option (it recovers regardless of why the ADSP is down), but if you'd prefer the dracut route I'm happy to rework it — or do both, since they address different halves.

jesherman:
Correction — my "fails on every boot" claim above was wrong, and the dracut angle I listed as an alternative is actually the real fix.

Those four boots were all running an initramfs I had rebuilt by hand while testing a custom kernel. Comparing it against the image-built one:

| initramfs | `q6v5` modules | qcom firmware |
|---|---|---|
| stock (image-built) | **0** | 0 |
| mine (hand-run dracut) | **2** | 0 |

The stock initramfs does not ship the ADSP driver at all, so it loads later from the rootfs where the firmware exists. Mine pulled `qcom_q6v5_pas.ko` in, so it probed at ~3s with no firmware available and failed deterministically. My own change turned an occasional failure into a guaranteed one, and I then measured the result and reported it as the baseline.

Rebuilding with `--omit-drivers 'qcom_q6v5_pas|qcom_q6v5'` restores stock behaviour:

```
adsp=running  battery=87        (no manual start)
"adsp is available" at ~7s, after switch-root
no request_firmware failure at all
```

What still stands: on stock images the ADSP does sometimes fail to come up, leaving the battery unreadable — boots -1 and -2 in my earlier logs were stock 7.0.11 and had no `charger_pd`. So the failure is real but **intermittent**, not the every-boot behaviour I described.

That makes this PR a safety net rather than the primary fix. Reasonable options:
- keep it as-is (recovers regardless of cause, cheap, no-op when healthy)
- close it and instead ensure the initramfs never carries `qcom_q6v5_pas`, so the probe cannot happen early
- both

I have no strong preference — happy to rework it either way. Flagging so the earlier framing does not mislead the review.

jesherman:
@virtudude I've not seen the 0 battery issue in latest builds so I think this is ok to close? Anything you still need from this PR?

justradical:
closing as requested, we may add firmwares to the initramfs through dracut if they're missing

SilkyShine:
i don't think this is dead, unfortunately; i'm hitting a similar problem on my Ayn Thor, tested on my own build of `20260824.6cd9496`. (which i also built *on* my Thor!) it seems to repro consistently when i build it myself, despite the pipeline succeeding. the **CDSP** firmware and **display/msm** firmware also got -2/ENOENT, though the msm seems to have retried until it worked. it's also not as cosmetic in my case, because desktop mode **black-screens** if the ADSP and CDSP aren't fixed first.

this was in the dmesg logs:

```
[    3.154786] remoteproc remoteproc0: adsp is available
[    3.154838] remoteproc remoteproc0: Direct firmware load for qcom/sm8550/ayn/thor/adsp.mbn failed with error -2
[    3.154842] remoteproc remoteproc0: powering up adsp
[    3.154859] remoteproc remoteproc0: Direct firmware load for qcom/sm8550/ayn/thor/adsp.mbn failed with error -2
[    3.154860] remoteproc remoteproc0: request_firmware failed: -2
[    3.165672] remoteproc remoteproc1: cdsp is available
[    3.166643] remoteproc remoteproc1: Direct firmware load for qcom/sm8550/ayn/cdsp.mbn failed with error -2
[    3.166652] remoteproc remoteproc1: powering up cdsp
[    3.166703] remoteproc remoteproc1: Direct firmware load for qcom/sm8550/ayn/cdsp.mbn failed with error -2
[    3.166708] remoteproc remoteproc1: request_firmware failed: -2
[    3.257002] msm_dpu ae01000.display-controller: Direct firmware load for qcom/a740_sqe.fw failed with error -2
[    3.554423] msm_dpu ae01000.display-controller: Direct firmware load for qcom/a740_sqe.fw failed with error -2
```

this resolves it until restart (and allows desktop mode to boot afterward):

```
sudo sh -c 'echo start > /sys/class/remoteproc/remoteproc0/state'
sudo sh -c 'echo start > /sys/class/remoteproc/remoteproc1/state'
```

i noticed that these modules are in my initramfs, but **not** the same commit built by your CI:

```
sudo lsinitrd /usr/lib/modules/7.2.0/initramfs.img | grep -E 'q6v5'
-rw-r--r--   1 root     root        16480 Jan  1  1970 usr/lib/modules/7.2.0/kernel/drivers/remoteproc/qcom_q6v5.ko
-rw-r--r--   1 root     root        72760 Jan  1  1970 usr/lib/modules/7.2.0/kernel/drivers/remoteproc/qcom_q6v5_pas.ko
```

i have no idea why they're here, i changed nothing about the pipeline... is it somehow pulling them in from the host or something, despite `--no-hostonly`? the easy solution probably *is* adding something like `--omit-drivers 'qcom_q6v5_pas|qcom_q6v5'` to the dracut args, but there may be a deeper mystery here as to why they're even in the initramfs at all...

jesherman:
you're only seeing it when you roll/build your own images, right? Since I closed this, I do occasionally see it but only when i roll my own builds. but tbh it's also easy to fix in those cases. not sure if it requires an upstream change?

SilkyShine:
it is my own builds, yes, but i would think that something is wrong with the build pipeline if that can happen unmodified, no? at the moment i'm just trying to get functional builds verbatim on my own hardware, and this is the only problem i'm seeing with doing so. i would consider it a fixable bug in the pipeline, even if it happens not to occur on CI.

Issue #23, opened 2026-06-21
Lossless Scaling / decky-lsfg-vk installation fails on ArmadaOS


Mokimiwa:
Permission denied:
/tmp/tmp7sbb980f/lib/liblsfg-vk.so

can't install decky LSFG-VK from decky store

virtudude:
@Mokimiwa Thanks for the report. I'll take a look. I know ROCKNIX has a forked version of the decky plugin so I'm guessing Armada might need something similar.

Mokimiwa:
@virtudude i completely installed armada on internal, when i check back on desktop mode the app armada internal install gone? if i want to remove armada on internal how i do that? do i need armada on sd card to acess this?

virtudude:
@Mokimiwa You need to boot Armada from the SD again, because you can't delete the partitions you're booted from. To boot from the SD again you need to delete the Armada boot partition from fastboot, because the ROCKNIX bootloader prefers internal over SD. To do that, hold VOL- while you're powering on the device and leave it in the bootloader. Then connect it to a PC and run `fastboot erase ROCKNIX` (the Armada boot partition is named for ROCKNIX so their bootloader finds it). Once you boot Armada from SD you can rerun the installer and choose to restore Android (which will factory reset it again due to the partition resize). 

Sorry there isn't an easier way right now. The ROCKNIX ABL preferring internal over SD makes it a bit tricky to uninstall.

Mokimiwa:
Oh i see, thank you so much for the detailed step by step info. i will further test it on my odin 2 mini and report back for issue, enhancement and features. goodjob for great update 🎉 looking forward cheers

xXJSONDeruloXx:
I am the author of the lsfg-vk decky plugin. It would need an update with a version of lsfg-vk which has arm arch compat path, and a few oddities for snapdragon chips. basically similar patches to lsfg-vk android but compiled in glibc rather than bionic

virtudude:
> I am the author of the lsfg-vk decky plugin. It would need an update with a version of lsfg-vk which has arm arch compat path, and a few oddities for snapdragon chips. basically similar patches to lsfg-vk android but compiled in glibc rather than bionic

@xXJSONDeruloXx would be helpful if i shipped lsfg-vk ARM in the OS? Or is it better to package it with the plugin?

xXJSONDeruloXx:
[...]

PR #215, opened 2026-08-04
Switch the Pocket EVO to real suspend to RAM

jesherman:
> [!IMPORTANT]
> **Do not merge before [armada-packages#23](https://github.com/armada-os/armada-packages/pull/23).** On a kernel without those patches, `mem` is the configuration that hangs the EVO on resume with a Steam session running — recoverable only by holding the power button.

## What changed

- `ayaneo-pocket-evo.conf`: `ARMADA_SUSPEND_MODE=mem`
- New `sleep.conf.d` drop-in pinning `SuspendState=mem` — systemd's default `mem standby freeze` falls through to s2idle on a failed deep suspend, which wedges SM8550 hard enough to need a power-button reset (we hit this during bring-up). ROCKNIX pins the same for this SoC.
- README: sleep section still described fake suspend as universal, stale since the Odin 3 moved to `mem`

## Results

Pocket EVO, on battery, Steam running: **27.8 mA vs 161.8 mA** standby — ~2 days to ~12. 1202s of uninterrupted deep suspend woken only by the RTC alarm, `success=3 fail=0`, session intact on resume.

## Gaps

Headline numbers are from a 7.0.11 build; packages has since moved to 7.1.5, confirmed working but only over a ~4 minute run. No multi-hour soak, single drain sample per mode, one device. The 5.8× gap is well beyond sampling error, but holding this until there's a longer soak on 7.1.5 is a reasonable call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,10 +176,15 @@ desktop. The **Bazaar** app store and the **Armada Installer**
 
 ### Power button and sleep
 
-Pressing the power button does a "fake suspend" (inspired by ROCKNIX) rather than
-real S3 sleep: it blanks the screen and freezes the session, and the same press
-wakes it. Because the device does not truly sleep, idle battery drain is higher
-than it would be with real suspend.
+What the power button does depends on the device.
+
+On the **AYN Odin 3** and **AYANEO Pocket EVO** it is a real suspend to RAM: the
+device genuinely powers down and the same press wakes it. Standby drain is much
+lower — measured at roughly a fifth of fake suspend on the Pocket EVO.
+
+Everywhere else it is a "fake suspend" (inspired by ROCKNIX): the screen blanks
+and the session freezes, but the device stays powered, so idle battery drain is
+higher than it would be with real suspend.
 
 ## Updating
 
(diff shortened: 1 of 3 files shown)


jesherman:
Marking this draft — found a blocker, and it's one my testing should have caught much earlier.

**The power button does not wake the device from deep suspend.** Pressing power suspends fine, then pressing it again does nothing; the device is genuinely asleep and unwakeable, which presents as a hang and needs a hold-power reset.

Every "successful" result I reported was woken by an **RTC alarm** (`rtcwake` or a `wakealarm` write). I never tested the actual user-facing wake path. That's a significant gap in the validation behind the numbers above.

Evidence, from a suspend with a 120s RTC backstop and a power press partway through:

```
pm_wakeup_irq=202                → pmic_arb RTC alarm
rtc@6100   wakeup count 1 → 2    → RTC fired
pwrkey     wakeup count 0 → 0    → power key registered nothing
```

`pmic_pwrkey` shows `wakeup = enabled` and `pon@1300:pwrkey` is registered as a wakeup source, so it is armed in principle — but it never fires from `deep`. The RTC proves PMIC-side IRQs *can* wake the SoC, so this looks specific to the pwrkey path rather than PMIC wake generally.

The kernel work in armada-packages#23 is unaffected — deep suspend itself holds fine and the drain numbers stand. It is only *this* PR, flipping the EVO over to `mem`, that must wait until power-key wake works.

Next step is finding why the pwrkey doesn't arm a PDC wake on SM8550. Leaving this in draft until then.

jesherman:
replaced by https://github.com/armada-os/armada/pull/228

Issue #14, opened 2026-06-15
Unable to download or select Proton 11.0 (ARM64) compatibility layer


antnyhills:
Problem Description
The ARM64 build variant of Proton 11.0 is completely missing from the runtime selection menus and cannot be downloaded or installed through the client interface on ARM64-based hardware.

Observed Behavior
Action: Navigating to the compatibility tool settings or searching the software library/tools repository for Proton 11.0 on an ARM64 system.

Result: The ARM64-specific build of Proton 11.0 does not appear in the available list of compatibility layers. Attempts to download from it's dedicated Steam install page results in a gray install icon that is unable to be selected.

Expected Behavior
Proton 11.0 (ARM64) should be visible, selectable, and downloadable as a global compatibility tool for ARM64 environments, matching the availability of the standard x86_64 runtime packages.

virtudude:
@antnyhills Thanks for the report. I'm aware of the issue you're referring to where Steam won't fetch other Proton versions if you select them in compatibility. FWIW the default that ships with Armada is the CachyOS build of Proton 11 ARM so you are using Proton 11 ARM, but I do want to figure out the issue you're seeing so people can switch between the CachyOS build and official.

astearon:
Using the CachyOS Proton11 does not seem to work as well as the normal ARM build of Proton 11, I cannot get Elite Dangerous to run under Armada but it is working under Rocknix for example.

virtudude:
Working on a fix for this. Registering CachyOS Proton 11 as the default compat tool is what's breaking downloading other versions of Proton. So I'll probably change things to ship the default Steam compat tool mappings and let people override as needed.

Raxcoms:
> Working on a fix for this. Registering CachyOS Proton 11 as the default compat tool is what's breaking downloading other versions of Proton. So I'll probably change things to ship the default Steam compat tool mappings and let people override as needed.

Proton 11 arm was missing entirely from the compatibility tools. It couldn't be selected in the dropdown inside the game properties

Trying to download it from the library would prompt me to stream it instead 

Forcing another compatibility layer to be the default in the developer settings allowed me do download proton 11 arm through the library but it still wouldn't appear in the selector, so I couldn't use it still

Is that something that will also be fixed by changing the default proton from the build? Or is it a different issue?

virtudude:
[...]

PR #220, opened 2026-08-05
Bump kernel for Pocket EVO GPIO buttons

jesherman:
Follow-up to #169.

Pins the kernel artifact built from armada-os/armada-packages#14. That kernel adds the missing Pocket EVO `aya-keys` device-tree node required by the InputPlumber mapping merged in #169. Without this bump, the userspace mapping is present but inert because the kernel exposes no matching input device.

Hardware validation on an AYANEO Pocket EVO:
- verified the published kernel artifact and extracted DTB contain `aya-keys`, GPIOs 85/91/169/170, and F13-F16
- booted the exact candidate DTB
- confirmed InputPlumber attached the GPIO input device
- observed one AYA SPACE press increment GPIO 85 and transition Steam from closed QAM to focused `Quick Settings`

This completes the kernel artifact step described in #169.


jesherman:
upon retesting im not sure if we still need this. will move to draft and re-open if i think its needed

jesherman:
Not needed as kernel was already bumped

Issue #191, opened 2026-08-01
Buttons Incorrectly Mapped


Mxndrot:
Follow up and splitting of issue ( #176 ) as requested by developers:

I’m having a few hardware button issues on the Pocket S 2K. First, the volume keys are completely inverted—the physical "+" button acts as "-" and vice versa. To enter the Rocknix/Armada ABL, you have to hold Volume + and Power, so the hardware mapping is different than other standard Android devices. This is fine, however, when you boot into the actual Armada OS on the device, unlike Android, the buttons do not map properly. In Android the device understands once booted what the volume + and Volume - should be and they respond properly, in armada they remain inverted.

Second, none of the extra buttons seem to be mappable. I can’t bind the LC, RC, or the two buttons on the bottom right; the system doesn’t seem to detect them at all or even know how to map them 

Third, USB peripherals are flaky or downright not working in desktop mode. My mouse and drawing tablet work fine in Android and on other PCs. in Armada, the mouse only works intermittently after several reboots, and the tablet isn’t detected at all, even after open tablet driver is installed. Bluetooth keyboards and mice work fine, so it appears isolated to the USB host controller or power management.

These issues all seem related so I’ve placed them all together as hardware support issues, please let me know if you have more questions about these issues

Macrey:
Confirming the inverted volume keys on my Pocket S 2K, with node attribution.

Pressing the physical "+" button three times, watching every system input node, the kernel reports KEY_VOLUMEDOWN each time:

    event1 (pmic_resin)  KEY_VOLUMEDOWN   <- physical "+"
    event2 (gpio-keys)   KEY_VOLUMEUP     <- physical "-"

The device tree explains it. The only gpio-keys node declares a single key:

    $ ls /proc/device-tree/gpio-keys/
    compatible  key-volume-up  name  pinctrl-0  pinctrl-names

That node is wired to the physical "-" button, while "+" arrives on the PMIC resin input, which is KEY_VOLUMEDOWN by Qualcomm convention. So the two are swapped relative to the hardware, in the DTS rather than anywhere in Armada which also matches the ABL entry combo being Volume + and Power.

Build: beta 20260729.d812101, compatible ayaneo,pocket-s1 qcom,sm8550.

On the second point in this issue (LC, RC and the two bottom-right buttons not being mappable): I've posted the measurements in #169, since it looks like the same missing gpio-keys node that PR addresses for the EVO.

[...]

PR #222, opened 2026-08-05
AYANEO Devices: ensure deck-uhid is selected

jesherman:
`controller-type` hardcodes `DEFAULT_TYPE = "deck-uhid"`, so every AYANEO gets it regardless of the `[xb360, keyboard]` its composite config asks for. InputPlumber applies the configured target at startup, then `armada-controller-type.service` overwrites it a second later over D-Bus:

```
Setting target devices: [xb360, keyboard]        <- 01-ayaneo-controller.yaml
Setting target devices: [deck-uhid, keyboard]    <- controller-type apply, ~1s later
```

`deck-uhid` presents to Steam as a Steam Deck Controller, which has a physical QAM button. Steam only offers its Guide+A → QAM fallback to controllers without one, so a device with no QuickAccess source of its own is left with no way into the Quick Access Menu at all.

Take the default from the device profile instead. `/etc/armada/controller.conf` and the Armada Control toggle still override it; an invalid or missing value falls back to `deck-uhid`.

- **Pocket EVO** — `deck-uhid`. AYA SPACE is a real QuickAccess button, and `=`, LC and RC are paddles, which only exist on that target.
- **Pocket S 2K** — `xb360`. Its device tree declares no `aya-keys` node and its MCU reports 11 buttons, so it emits neither BTN5 (QuickAccess) nor any paddle button in `ayaneo_mcu_xbox.yaml`. Nothing live is lost, and Guide+A works again. Should move back to `deck-uhid` when it gets an `aya-keys` node, since paddles only exist there.
- Everything else keeps `deck-uhid` via `defaults.conf`.

**Pocket DS** is deliberately left alone: if its MCU emits BTN5 it already has QAM today, and `xb360` would risk its paddles.

## Measured on a Pocket EVO

| Target | AYA SPACE → QAM | Guide+A | Guide+X | paddles |
|---|---|---|---|---|
| `deck-uhid` | works | dead | dead | work |
| `xb360` | works | opens QAM | opens menu | dead |

QuickAccess reaches Steam on both targets, so the paddles decide the split. This is not caused by advertising the capability — with every QuickAccess mapping removed, Guide+A still did nothing on `deck-uhid`, so #75 is not implicated.

## Testing

S 2K owners can reproduce without a build: `sudo /usr/libexec/armada/controller-type set xb360`. DS owners, check whether your AYA button and paddles work today — that decides whether the DS should join.

Addresses the missing QAM half of #191 / #50 with no device-tree change.

Guide+X is unaffected either way: `SHOW_KEYBOARD` never fires on ARM, and `xb360` loads no chord file. The README known-issue entry is updated to say so.

### DIFF
--- a/system_files/usr/lib/armada/devices/ayaneo-pocket-evo.conf
+++ b/system_files/usr/lib/armada/devices/ayaneo-pocket-evo.conf
@@ -4,3 +4,5 @@ ARMADA_SOC_CLASS=SM8550
 
 ARMADA_PANEL_ORIENTATION=left
 ARMADA_GAMEPAD_QUIRK=ayaneo-xbox
+# AYA SPACE and the =/LC/RC paddles only exist on deck-uhid; Guide+A is the cost.
+ARMADA_CONTROLLER_TYPE=deck-uhid
--- a/system_files/usr/lib/armada/devices/ayaneo-pocket-s2k.conf
+++ b/system_files/usr/lib/armada/devices/ayaneo-pocket-s2k.conf
@@ -4,3 +4,6 @@ ARMADA_SOC_CLASS=SM8550
 
 ARMADA_PANEL_ORIENTATION=right
 ARMADA_GAMEPAD_QUIRK=ayaneo-xbox
+# No aya-keys node, so no QAM/paddle buttons reach userspace: xb360 restores
+# Guide+A with nothing live to lose. Move back to deck-uhid once it gets aya-keys.
+ARMADA_CONTROLLER_TYPE=xb360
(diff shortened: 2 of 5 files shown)


JPyke3 (changes requested):
Could you please remove the amount of comments in this PR

jesherman:
done, thanks for the feedback!

JPyke3:
This one for sure needs a rebase - So I'd recommend doing that.

But this one is also likely best to sit with @justradical

justradical:
I am tackling this in a different way that will affect all devices, i might need your help with these ones that i dont have at a little later point

justradical:
This isnt needed anymore to my knowledge, lmk or re-open if im incorrect

PR #223, opened 2026-08-05
Add config-driven power profiles with real-frequency controls

aumVilohm:
## What

Expand the power system from three fixed profiles to a config-driven set, shipped with seven, because more tiers let powerful systems be tuned more precisely. Armada Control gains real-frequency controls.

## Changes

**armada-powerd**
- The profile list is discovered from the `[profile.*]` sections, with an optional `[general] profile_order`, instead of a hardcoded eco/balanced/performance.
- `profile_settings` falls back to the default profile, then the first defined profile.

**Armada Control (Power tab)**
- Lists every configured profile.
- CPU max frequency set per cluster in real MHz (efficiency and performance cores), from the available cpufreq steps.
- GPU minimum and maximum set in real MHz from the OPP list.
- Per-profile CPU governor selector.
- The backend ships the available CPU frequencies, GPU OPPs, and governors.

**Config**
- `power-profiles.conf` now defines a seven-tier ladder tuned for SM8750 (custom underclock levels and fan curves). The SM8550 and SM8650 underclock tables are kept intact; those SoCs would need analogous extra tiers to use the full ladder, otherwise the extra tiers fall back to top frequency there.

## Testing

Built and run on a KONKR Pocket FIT Elite (SM8750). All seven profiles switch and apply, per-cluster MHz and governor take effect, and the fan curves match the tuned values. Daemon config parsing and the plugin save path were checked with a config round-trip.

### DIFF
--- a/decky/armada-control/py_modules/armada_control/config.py
+++ b/decky/armada-control/py_modules/armada_control/config.py
@@ -1,5 +1,5 @@
 from .controller import CONTROLLER_TYPES, controller_type
-from .power import factory_power_defaults, parse_power
+from .power import available_hardware, factory_power_defaults, parse_power
 from .steam import installed_games
 from .system import abl_version, cpu_device_class, os_version, ssh_enabled
 from .tweaks import fex_profile_labels, load_fex_contract, load_tweaks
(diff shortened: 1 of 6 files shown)


aumVilohm:
Superseded by #226, which is rebased on current `main` (with the daemon reconciled against the new per-game performance logic) and adds a graphical fan-curve editor. Closing this one in favor of #226.

PR #227, opened 2026-08-06
Add a sing-box VPN toggle, profiles, and a browser-based config import

aumVilohm:
## Summary

Adds an optional sing-box VPN section to Armada Control: an enable toggle, two profiles, and a browser-based config import.

## What's included

- **VPN toggle and two profiles** in the Advanced tab: enable/disable the sing-box service and switch between `config-1` and `config-2`.
- **Import VPN config**: a button that opens a small local web server on the device, shown as a QR code and URL in the QAM. Open it from a computer or phone on the same network and paste a `vless://` link or a full sing-box `config.json`. The page has two buttons (Save to Profile 1 / Save to Profile 2). The device validates the config with `sing-box check` and writes it to the chosen profile, restarting the service if that profile is active.
- **vless:// conversion**: a pasted share link is converted into a full sing-box config, pinning the server to an IP and reusing the profile's existing config as a template so its dns, tun and route settings are preserved.
- **sing-box install, service, and helper**: a pinned sing-box binary installer, a systemd service, and a small `sauna-vpn` helper.

## Notes

- The import web server is HTTP, ephemeral (it runs only while the import dialog is open), and write-only.
- User VPN configs are supplied by the user and are never bundled in the image.
- Adds `qrcode.react` for the QR code.
- Touches `config.py`, `types.ts` and `package.json`, which also change in the power-profiles PR; whichever merges second needs a trivial rebase.

## Testing

Validated on device (KONKR / AYANEO Pocket FIT Elite, SM8750): the toggle, the profile switch, and the import flow end to end (paste a `vless://` link in a computer browser, the config is written, and the VPN connects and routes traffic).

### DIFF
--- a/build_files/46-install-singbox.sh
+++ b/build_files/46-install-singbox.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+set -euxo pipefail
+
+# sing-box static binary, pinned release + sha256 (arm64).
+SB_VER=1.13.16
+SB_SHA=d587fb00bdc3c044227f35d15d154f271bc75108475091eda2542e4b82bb2949
+curl -fsSL -o /tmp/sing-box.tar.gz     "https://github.com/SagerNet/sing-box/releases/download/v${SB_VER}/sing-box-${SB_VER}-linux-arm64.tar.gz"
+echo "${SB_SHA}  /tmp/sing-box.tar.gz" | sha256sum -c -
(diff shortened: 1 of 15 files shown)


justradical:
Hey, thanks for taking the time to try to contribute but sadly this will not be merged into Armada, you might be able to make your own decky plugin and get it added to armada store or to the normal decky loader store, this is way outside of our scope as a project so I'm closing this PR

Issue #28, opened 2026-06-22
AYN Thor Second Screen


panpanpanpaan:
Even if the second screen it's not integrated, it would be cool to be able to use it in the desktop mode to play around, is it something that can be done? thank you"

virtudude:
> Even if the second screen it's not integrated, it would be cool to be able to use it in the desktop mode to play around, is it something that can be done? thank you"

I had to disable the lower screen at the kernel level because Steam was finding it for brightness control instead of the upper screen. It was a quick workaround and my hope is to find a way to patch gamescope or whatever system is involved in choosing the primary panel so I can bring it back.

xXJSONDeruloXx:
Potentially relevant https://gist.github.com/NaviVani-dev/9a8a704a31313fd5ed5fa68babf7bc3a

SalamanderThing:
> I had to disable the lower screen at the kernel level because Steam was finding it for brightness control instead of the upper screen. It was a quick workaround and my hope is to find a way to patch gamescope or whatever system is involved in choosing the primary panel so I can bring it back.

would it be possible to somehow use the second screen as a trackpad?

brycesub:
Dug into this on real hardware — PRs are up: virtudude/armada-packages#1 (kernel/DT) and #78 (userspace).

**Root cause of the disable:** with both panels enabled, the bottom panel's backlight (`ae94000.dsi.0`, range 0-255) enumerates before the top one (`ae96000.dsi.0`, range 0-4096) and Steam grabs the first backlight it finds for brightness control. The panels themselves work fine on the ROCKNIX device tree — `panel-ddic-ch13726a` for the bottom screen is already in the Armada kernel.

**What the PRs deliver (phase 1):**
- Gaming mode: unchanged — Steam on the top screen; brightness writes are steered to the top panel in `steamos-priv-write` (with range rescaling); the bottom screen stays dark and its touchscreen is inhibited so resting fingers don't move the cursor
- Desktop mode: one KDE Wayland session spanning both screens, top primary with the bottom centered below, correct rotation, and each touchscreen pinned to its own screen

**Why gaming mode can't get a desktop on the bottom screen yet (phase 2):** gamescope in embedded mode is the exclusive DRM master of the GPU — no other compositor can drive the second connector while it runs, and gamescope's DRM backend only lights up one output. Getting Steam-on-top + desktop-on-bottom simultaneously needs one of:
[...]

PR #235, opened 2026-08-10
feat(fake-suspend): opt-in radio suspend to cut idle drain

unjordi:
## What & why

Under `fake-suspend`, the Wi-Fi/BT radios stay powered and are a large share of the idle drain. The `suspend_radios`/`resume_radios` helpers already exist but their calls were commented out, with the note *"disabled until we can improve resume speed (currently takes 30+ seconds)"*.

Rather than re-enable them globally (and re-introduce that resume regression for everyone), this makes it an **opt-in per device**, defaulting to today's behavior.

## Measurements (Retroid Pocket 6, SM8550)

Device otherwise idle in `fake-suspend`, battery at ~97%, sampled `current_now` every 2s:

| Radios | Avg draw |
|---|---|
| ON | ~300 mA |
| OFF (`rfkill` Wi-Fi+BT) | ~180 mA |
| ON again (control) | ~273 mA |

→ radios ≈ **~40%** of fake-suspend idle drain (~120 mA). On a ~5000 mAh pack that's roughly **~17h → ~28h** of standby.

Resume cost, measured on the same RP6: **~17s** from `rfkill unblock` to Wi-Fi connectivity restored (BT/app-level re-init may add a bit more).

## The change

- `devices/defaults.conf`: document + default `ARMADA_SUSPEND_RADIOS=0`.
- `libexec/armada/device-env`: export `ARMADA_SUSPEND_RADIOS`.
- `libexec/armada/fake-suspend`: gate the existing `suspend_radios`/`resume_radios` calls on `ARMADA_SUSPEND_RADIOS=1`.

**No device is opted in** by this PR — behavior is unchanged unless a device's `.conf` sets `ARMADA_SUSPEND_RADIOS=1`. Happy to also flip it on for `retroid-pocket-6.conf` in this PR if you'd like it on by default there, given the numbers above — your call on the resume-latency tradeoff.


unjordi:
Hey! 👋 First — thank you for ArmadaOS. It turned my Retroid Pocket 6 into a proper Deck-mini and I've been settinng it up for weeks (and playing for about ... 6 hrs total xD)

BUT, in my defense! The stock android never even saw daylight (wifi) before the sd was backed up and armada was flashed in. 

WhatsApp Image 2026-07-28 at 5 30 15 PM

aaanyway, small offering here. I kept bleeding battery to fake-suspend whenever I'd put the RP6 to sleep and toss it in a bag, so I (mr Claude) poked at the radios. On my unit (SM8550) they're about 40% of the idle drain — ~300 mA dropped to ~180 mA with Wi-Fi/BT soft-blocked (measured live over ssh). 

This just makes the existing suspend_radios/resume_radios helpers reachable again behind an opt-in ARMADA_SUSPEND_RADIOS flag, default off so nothing changes for anyone unless a device opts in. Resume cost on my RP6 was ~17s to Wi-Fi, so I get why it was parked — hence opt-in rather than on-by-default. Totally your call whether any device flips it on (happy to add it to retroid-pocket-6.conf if you want it).

 Unrelated and zero pressure: I noticed the deep-suspend stack (#28/#29) was all measured on a Pocket EVO. I've got a real RP6 right here and I'm more than happy to be a guinea pig if it'd help validate the SM8550 bring-up on actual RP6 silicon — drain numbers, resume behavior, whatever's useful. Just say the word. 🙌

DAmmit claude! did you have to tell them about the "guinea pig" propossition?!m THAT WAS BETWEEN YOU AND ME! 

bUT YEAH, Thanks again for building this!

best regards: a human, I think.

JPyke3 (changes requested):
Hey there!

I do appreciate this PR and it's a good one for fake suspend. The only thing I might suggest is probably reducing how verbose the line comments are.

Claude code has a nasty habit of adding huuuge comments - this one is no exception.

With that one tidied up I reckon this could work!

JPyke3:
Also if you could fix the conflicts :)

PR #237, opened 2026-08-10
Start Steam reliably in Desktop Mode

Janleyx:
## What changed

- Add a KDE autostart entry for Steam Desktop Controls.
- Wait for Plasma Shell and Xwayland before launching the ARM Steam client.
- Import Plasma’s live display environment before startup.
- Avoid a second Steam instance and skip startup inside Nested Desktop.

## Why

Starting Steam in full Desktop Mode enables Steam Input/controller desktop controls, but starting it too early can crash the ARM client’s legacy UI initialization. This keeps the behavior independent from Nested Desktop so it can be reviewed and tested separately.

## Validation

- `bash -n system_files/usr/libexec/armada/desktop-steam-autostart`
- `desktop-file-validate system_files/etc/xdg/autostart/armada-desktop-steam.desktop`
- `git diff --check`
- Full Armada shell test suite passed on the current upstream base.

### DIFF
--- a/system_files/etc/xdg/autostart/armada-desktop-steam.desktop
+++ b/system_files/etc/xdg/autostart/armada-desktop-steam.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=Steam Desktop Controls
+Comment=Start Steam to provide gamepad controls in Desktop Mode
+Exec=/usr/libexec/armada/desktop-steam-autostart
+Terminal=false
+OnlyShowIn=KDE;
+X-KDE-autostart-after=panel
+X-KDE-StartupNotify=false
--- a/system_files/usr/libexec/armada/desktop-steam-autostart
+++ b/system_files/usr/libexec/armada/desktop-steam-autostart
@@ -1,0 +1,47 @@
+#!/usr/bin/bash
+set -euo pipefail
+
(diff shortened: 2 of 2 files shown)


justradical:
Has this script been tested in Armada? Does it really need all this logic to be able to launch through xdg? Steam already detects if its running so why are you also checking that?

Janleyx:
Sorry theres 1 thing here thats actually supposed to be tied to my nested desktop pr. I must of overlooked it. Import Plasma’s live display environment before startup is intended for the nested desktop work. The reason behind it when you launch nested desktop in Game mode the desktop environment loads in a vertical alignment. So it only takes half the screen and to correct it you have to rotate the screen in nested desktop. The problem however is when you go back to desktop mode you're going to find the screen is vertical now because of the change that was done in nested desktop. So this correction in particular is to avoid that from happening both in desktop mode and in nested desktop they both retain the correct screen orientation.

Janleyx:
I was pretty busy theese days and did the pr seperation request in a bit of a hurry. My bad.

Janleyx:
Apologies for the long delay here. I have been very busy lately, but I plan to revisit this patch soon, update it against the latest Armada release, and revalidate it on-device before requesting another review. Thank you for your patience.

PR #244, opened 2026-08-13
fix: enable sync suspend on Pocket DS

pacoa-kdbg:
## Summary
- Add a per-device `ARMADA_PM_SYNC=1` quirk exported through `device-env`.
- Apply the quirk from `armada-device-quirks` by disabling kernel async suspend (`/sys/power/pm_async=0`) when requested.
- Enable sync PM only for AYANEO Pocket DS to avoid a Goodix touchscreen suspend ordering race.

## Why
On AYANEO Pocket DS, deep sleep with async PM enabled intermittently aborts during suspend:

```text
Goodix-TS 2-005d: PM: dpm_run_callback(): goodix_suspend [goodix_ts] returns -11
Goodix-TS 2-005d: PM: failed to suspend async: error -11
PM: Some devices failed to suspend, or early wake event detected
```

Using synchronous PM ordering lets the device complete suspend. Subsequent wakes are reported as normal PMIC power-key wakeups.

## Test plan
- [x] `bash -n system_files/usr/libexec/armada/device-env system_files/usr/libexec/armada/device-quirks tests/perf-settings-test.sh`
- [x] Verified `device-env` emits `ARMADA_PM_SYNC=1` for `ARMADA_MODEL='AYANEO Pocket DS'`.
- [x] Live Pocket DS: installed a temporary systemd drop-in with `Environment=ARMADA_PM_SYNC=1` and `ExecStartPre=/bin/sh -c 'echo 0 > /sys/power/pm_async'`; verified `/sys/power/pm_async=0`.
- [x] Hardware: AYANEO Pocket DS, kernel `7.2.0-rc7`: with sync PM (`pm_async=0`), deep suspend completed successfully and resumed via `pmic_pwrkey`.
- [ ] `tests/perf-settings-test.sh` on this local ARM runner: blocked by existing host limitation (`unknown cpus: [4, 5, 6, 7]`) because the runner exposes only CPUs `0-3` while the test assumes an 8-core topology.

## Scope
This is Pocket DS-exclusive. It avoids changing suspend behavior for tested devices until they have their own evidence-backed device quirk.


pacoa-kdbg:
I did about 50 iterations of sleep.- wake and works as expected

PR #246, opened 2026-08-13
Add proton downloader from ChimeraOS

justradical:
Downloads GE-Proton and Proton-CachyOS in the background when trying to start a game using one of them instead of shipping them with the image

Currently there's some dead code in there referring to the chimera-splash, this can either be removed or replaced with our own splash


justradical:
This is currently blocked by the frame release because we dont want to push Proton CachyOS out of the image until we are sure that Valve wont randomly remove access to the built-in Proton versions

PR #247, opened 2026-08-14
Update system FEX layout and remove Proton registration / require_tool_appid stripping

virtudude:
- Current Steam chains its own FEX depot under every x86 title and needs an OS mounted rootfs at /usr/share/guestos/fex-mesa. armada-guestos.service provides the mount. mesa-x86 package layers armada-patched x86 Mesa into the tree.
- Per-game FEX config moves to ~/.cache (visible in pressure-vessel).
- Dead EGL thunk toggle removed.
- Arch RootFS bumped to 2026-08-11.
- Stop self-registering Proton 11 ARM and stripping require_tool_appid. Add cleanup of the old registration.

### DIFF
--- a/build_files/30-install-steam-session.sh
+++ b/build_files/30-install-steam-session.sh
@@ -48,16 +48,22 @@ dnf5 -y install --setopt=install_weak_deps=False /packages/fex/fex-emu-*.rpm
 
 # Use Arch rootfs for better compatibility with Linux games targeting SteamOS
 mkdir -p /usr/share/fex-emu/RootFS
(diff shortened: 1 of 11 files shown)


build_files/30-install-steam-session.sh
JPyke3, line 126: Looks like `launch-steam` is no longer stripping `require_tool_appid`, but we do have another sed here. Might be worth cleaning this too?

JPyke3 (dismissed):
Looks all good - I haven't tested on device yet but this is the only thing I thought we might've missed.

PR #278, opened 2026-08-18
fix(fixups): resolve LSFG-VK Vulkan layer library and manifest paths

drewano:
## Description

Ensures the `liblsfg-vk.so` Vulkan layer library and its implicit layer manifest are properly resolved when `Decky LSFG-VK` is installed.

### Context & Problem
- When installing the `Decky LSFG-VK` plugin from the Armada Store / Decky, the upstream installer writes a relative library path (`"../../../lib/liblsfg-vk.so"`) into `~/.local/share/vulkan/implicit_layer.d/VkLayer_LS_frame_generation.json`.
- Because this relative path cannot be resolved by the Vulkan loader, `vulkaninfo` and games fail to load the layer with:
  ```text
  ERROR: [Loader Message] Code 0 : ~/.local/share/vulkan/implicit_layer.d/../../../lib/liblsfg-vk.so: cannot open shared object file: No such file or directory
  ```
- As a result, frame generation silently fails to hook into the game.

### Solution
- `armada-fixups` checks if the `Decky LSFG-VK` ARM64 binary exists and ensures `~/.local/lib/liblsfg-vk.so` is installed.
- Normalizes `VkLayer_LS_frame_generation.json`'s `library_path` to the absolute path `~/.local/lib/liblsfg-vk.so`.

## Testing
- Verified on AYN Odin 2 (Snapdragon 8 Gen 2 / SM8550) with `vulkaninfo` and *Mortal Sin* via `~/lsfg %command%`.
- Vulkan loader successfully initializes `VK_LAYER_LS_frame_generation`.

### DIFF
--- a/system_files/usr/libexec/armada/armada-fixups
+++ b/system_files/usr/libexec/armada/armada-fixups
@@ -1,7 +1,9 @@
 #!/usr/bin/env python3
+import json
 import os
 import pathlib
 import re
+import shutil
 import stat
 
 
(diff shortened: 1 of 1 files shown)


justradical:
Would this be better solved as a patch to the plugin itself? The dev is pretty responsive on PRs i think

drewano:
> Would this be better solved as a patch to the plugin itself? The dev is pretty responsive on PRs i think

Yessss Ok I will do this 🙌

PR #286, opened 2026-08-20
Add Bug Report and Feature Request issue templates

Raxcoms:
Added these two templates for issues

|  **Bug Report**  | **Feature Request** |
| ------------- | ------------- |
| image  | image |


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
+name: "Bug report"
+description: "Report a bug in Armada."
+title: "[Bug]: A short description of your bug"
+labels: ["Bug"]
```
justradical: This should be `type: bug` and not `labels: ["Bug"]`, this is probably an organization-only feature so you just have to manually edit it for this, same goes for feature requests, the default feature type should be called feature so there it's going to be `type: feature`

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
+name: "Bug report"
+description: "Report a bug in Armada."
+title: "[Bug]: A short description of your bug"
```
justradical: title shouldn't be prefixed on either of these, its inferred from the type so this is unnecessary noise

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
...
+assignees: []
+body:
+  - type: input
+    id: hardware
+    attributes:
+      label: Hardware Device
+      description: Please enter the device you are using.
+      placeholder: e.g. AYN Thor
+    validations:
+      required: true
```
justradical: is it possible to make this a dropdown that adds labels automatically? We have specific device labels set up for issues. This also goes for feature requests

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,67 @@
...
+    validations:
+      required: false
+
+  - type: textarea
+    id: comments
+    attributes:
+      label: Additional comments
+      description: Any other context, rationale, or constraints that will help evaluate or implement this feature.
+    validations:
+      required: false
```
justradical, line 64: is there a reason for 3 separate text areas here? I can see maybe the comments but desired behavior and description are basically the same for this template

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,87 @@
...
+
+  - type: textarea
+    id: steps
+    attributes:
+      label: Steps to reproduce
+      description: Include exact commands, menu clicks, or actions. If applicable, attach logs or screenshots.
+      placeholder: |
+        1. Step 1
+        2. Step 2
+        3. Step 3
```
justradical: would probably be helpful with some examples here instead of just step 1, step 2, step 3
like:
```
Example:
1. Open QAM
2. Go to Armada Control
3. X thing happened when i expected Y
```

justradical (changes requested):
Looks really nice but a couple of things here need changing, if you have any notes or comments for any of these then just lmk

.github/ISSUE_TEMPLATE/bug_report.yml
justradical: Does title include a placeholder? If not then maybe it should be blank? I'm not sure

Raxcoms:
@justradical I made the changes you suggested
The type property breaks the templates on my repo since I don't have those, so I can't verify if that is working
Eveything else should be fine

justradical (approved):

justradical:
> @justradical I made the changes you suggested The type property breaks the templates on my repo since I don't have those, so I can't verify if that is working Eveything else should be fine

guess we'll find out

Issue #40, opened 2026-06-30
Armada Control not applying profile settings


iaics:
Hi! I've just discovered this project and I'm excited to help testing. Also thank you for the quick fix on the unreadable SD Card problem!

Anyway, Armada Control has a few issues with the Power Profile settings. 

It completely refuses to change the power profile from Balanced to other profiles. Even if it lets you change it, as soon as you switch to another tab or close the side menu, it resets to Balanced.

The Fan Curve doesn't have this behaviour since it displays what you select, but it doesn't work. The fan doesn't ramp up and it stays at the same constant speed no matter what option you choose.

I suspect the same thing could be happening with the underclock, but I didn't test that.

I'm using an Odin 2 Portal Base with a Sandisk Extreme Pro 256Gb. 
Keep in mind that I'm running the OS from the SD Card, I didn't install it on the internal storage.

If you need anything like logs, let me know where I can find them, and I would gladly post them here.

Thank you and keep up the great work!

virtudude:
> Hi! I've just discovered this project and I'm excited to help testing. Also thank you for the quick fix on the unreadable SD Card problem!
> 
> Anyway, Armada Control has a few issues with the Power Profile settings. 
> 
> It completely refuses to change the power profile from Balanced to other profiles. Even if it lets you change it, as soon as you switch to another tab or close the side menu, it resets to Balanced.
> 
> The Fan Curve doesn't have this behaviour since it displays what you select, but it doesn't work. The fan doesn't ramp up and it stays at the same constant speed no matter what option you choose.
> 
> I suspect the same thing could be happening with the underclock, but I didn't test that.
> 
> I'm using an Odin 2 Portal Base with a Sandisk Extreme Pro 256Gb. 
> Keep in mind that I'm running the OS from the SD Card, I didn't install it on the internal storage.
> 
> If you need anything like logs, let me know where I can find them, and I would gladly post them here.
> 
> Thank you and keep up the great work!

Thanks for reporting this. I suspect the UI might be confusing based on your description. The power profiles section in Armada Control is for editing the profiles, not changing them. You change the current profile in the performance section of the Quick Access Menu. I did that so the power profile could be part of the per game settings. Does that explain the behavior you're seeing? If so I'll revisit the Armada Control UI. I was worried it might be confusing when I designed it.

iaics:
Ok now that makes sense ahahahahaahah. I feel stupid now.

[...]

PR #323, opened 2026-08-26
feat(inputplumber): complete KONKR Pocket FIT controller support

mydanyi:
## Background

Armada already lists the KONKR Pocket FIT as a supported device and ships an InputPlumber profile for it, so I am not adding a new device from scratch.

While testing the existing support on the physical device, I found that the profile only covered part of the controller. The main gamepad, auxiliary button device, and controller modes were not combined correctly. Some standard and device-specific buttons were unavailable, the four back buttons were not exposed independently, and AYANEO/DInput mode had no rumble. An early rumble implementation also drove only one motor.

I completed the controller profile in this PR. The matching InputPlumber hidraw rumble driver is provided by [armada-packages#40](https://github.com/armada-os/armada-packages/pull/40).

## Controller mapping

I combined the `4001:0428` AYANEO/DInput main gamepad and the `1c4f:0002` AYANEO auxiliary button device into one InputPlumber composite controller. This prevents Steam from opening the raw gamepad alongside InputPlumber's virtual controller.

The standard controls are mapped as follows:

| Physical input | Virtual controller input |
| --- | --- |
| `BTN_SOUTH` | A / South |
| `BTN_EAST` | B / East |
| `BTN_WEST` | X / West |
| `BTN_NORTH` | Y / North |
| `BTN_START` | Menu / Start |
| `BTN_SELECT` | View / Select |
| `BTN_MODE` | Steam / Guide |
| `BTN_TL`, `BTN_TR` | LB, RB |
| `ABS_BRAKE`, `ABS_GAS` | LT, RT |
| `ABS_X/Y` | Left stick |
| `ABS_Z/RZ` | Right stick |
| `BTN_THUMBL`, `BTN_THUMBR` | L3, R3 |
| `ABS_HAT0X/Y` | D-pad |

The device-specific controls are mapped as follows:

| Physical control | Raw input | Final action |
| --- | --- | --- |
| Side shortcut that was previously mapped to Steam+A | `BTN7` | Quick Access Menu |
| Front shortcut that previously opened QAM | `BTN6` | Steam+X / virtual keyboard |
| Second AYANEO shortcut | `BTN5` | Screenshot |
| Upper-left back button | `BTN3` | L4 |
| Lower-left back button | `BTN_Z` | L5 |
| Upper-right back button | `BTN4` | R4 |
| Lower-right back button | `BTN_C` | R5 |

This removes the incorrect Steam+A action and prevents both back buttons on each side from collapsing into the same logical button.

I changed the default virtual target to the Steam Deck controller while keeping the keyboard and mouse targets.

## Rumble

I kept the physical controller in AYANEO/DInput mode so all four back buttons remain available instead of forcing Xbox mode as a workaround.

This profile adds the `4001:0428` interface 0 hidraw source used by the driver in armada-packages#40. The controller uses an 8-byte HID output report. Physical probing confirmed that byte 4 controls the left motor and byte 5 controls the right motor; the earlier one-sided result was caused by a one-byte offset.

## Tested

I tested this on:

- KONKR Pocket FIT
- Snapdragon 8 Gen 3 / SM8650
- Armada
- InputPlumber 0.77.2
- Steam Deck virtual controller target
- Native ARM64 release build

Automated validation:

- AYANEO rumble unit tests: 3 passed
- InputPlumber library tests: 13 passed
- Armada controller profile test passed
- Rust formatting check passed
- ARM64 Docker/QEMU release build passed
- Target replacement and rumble reset passed
- Left-only and right-only rumble passed through the real InputPlumber force-feedback path

I also verified every control on the physical device:

- A/B/X/Y, D-pad, bumpers, triggers, sticks, and stick clicks work;
- Steam, QAM, virtual keyboard, and screenshot shortcuts work;
- the four back buttons report independently as L4, L5, R4, and R5;
- the left and right motors work independently;
- normal in-game rumble works;
- InputPlumber remains active without crashes or service restarts.

Disclosure: I used OpenAI Codex to assist with analysis, implementation, and testing. I reviewed the final diff and validated it on the physical device.

### DIFF
--- a/system_files/usr/share/inputplumber/capability_maps/ayaneo_type9.yaml
+++ b/system_files/usr/share/inputplumber/capability_maps/ayaneo_type9.yaml
@@ -1,0 +1,61 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ShadowBlip/InputPlumber/main/rootfs/usr/share/inputplumber/schema/capability_map_v2.json
+version: 2
+kind: CapabilityMap
+name: AYANEO Type 9
+id: aya9
+
+mapping:
+  - name: Side button to Quick Access
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN7
+          value_type: button
+    target_event:
+      gamepad:
+        button: QuickAccess
+  - name: Front button to virtual keyboard (Steam)
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN6
+          value_type: button
+    target_event:
+      gamepad:
+        button: Guide
+  - name: Front button to virtual keyboard (X)
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN6
+          value_type: button
+    target_event:
+      gamepad:
+        button: North
+  - name: AYANEO button 2 to screenshot
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN5
+          value_type: button
+    target_event:
+      gamepad:
+        button: Screenshot
+  - name: Upper left back button
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN3
+          value_type: button
+    target_event:
+      gamepad:
+        button: LeftTop
+  - name: Upper right back button
+    source_events:
+      - evdev:
+          event_type: KEY
+          event_code: BTN4
+          value_type: button
+    target_event:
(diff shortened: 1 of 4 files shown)


mydanyi:
I have fixed this in the upstream PR, so I am closing this PR: https://github.com/ShadowBlip/InputPlumber/pull/670

PR #325, opened 2026-08-26
feat(armada-control): add Simplified Chinese localization

mydanyi:
## Summary

I use Armada Control with Steam set to Simplified Chinese, but the plugin currently keeps its own UI in English. This also affects newer features such as the fan curve editor and RGB controls.

This PR adds Simplified Chinese localization while keeping English as the default:

- follows the Steam client language first, then the Decky and browser locales;
- translates Compatibility, Power, Fans, Settings, calibration, dialogs, the full-screen fan curve editor, and RGB controls;
- leaves unknown runtime labels unchanged;
- falls back to English for unsupported languages;
- adds tests for locale detection, precedence, interpolation, runtime labels, and English fallback.

English strings remain the source keys, so the existing English interface is not replaced. `schinese`, `SteamChina_SChinese`, `zh-CN`, `zh-Hans`, and `zh-SG` are mapped to Simplified Chinese. Traditional Chinese is not included and currently falls back to English.

### Automated checks

The ARM64 production bundle was built and checked with:

- `npm ci`
- `npm run test:i18n`
- `npx tsc --noEmit`
- `npm run build`

### Physical device testing

I tested this build on an actual physical handheld, not only in a container or emulator:

- Device: KONKR Pocket FIT (G3 Gen 3)
- SoC: Qualcomm Snapdragon 8 Gen 3
- Architecture: ARM64
- OS: ArmadaOS `20260825.b31e234`
- Steam language: Simplified Chinese

The physical-device test passed:

- Compatibility, Power, Fans, and Advanced pages display correctly in Simplified Chinese;
- the newly added fan controls and fan responsiveness settings are translated;
- the full-screen fan curve editor opens, displays, and closes correctly;
- the plugin loads normally and no Armada Control backend errors appear in the PluginLoader logs;
- the interface was manually reviewed and accepted on the physical device.

RGB controls are not exposed by this particular device, so their localization was covered by the type check and production build rather than live RGB hardware interaction.


decky/armada-control/src/i18n.ts
```diff
@@ -0,0 +1,283 @@
...
+  "The 0% minimum applies globally while Fan Stop is enabled.": "启用风扇停转期间，全局最低转速为 0%。",
+  "Fullscreen Editor": "全屏编辑器",
+  "Hide Points": "隐藏节点",
+  "Edit Curve Points": "编辑曲线节点",
+  "Add Point": "添加节点",
+  "Temperature (°C)": "温度 (°C)",
+  "Edit Point": "编辑节点",
+  "Stop Editing": "停止编辑",
+  "FAN STOPPED": "风扇已停转",
+  "D-Pad moves point {current} of {total} · LB/RB switches points · B stops": "方向键移动第 {current}/{total} 个节点 · LB/RB 切换节点 · B 停止编辑",
```
tslmy: “total个中第current个节点”听起来可能比”current/total个节点”更idiomatic一些。

decky/armada-control/src/i18n.ts
```diff
@@ -0,0 +1,283 @@
...
+  "Resetting...": "正在重置……",
+  "Armada Fans": "Armada 风扇",
+  "SAVE": "保存",
+  "Saving...": "正在保存……",
+  "Save Changes": "保存更改",
+  "Revert Changes": "撤销更改",
+  "You have unsaved changes.": "有尚未保存的更改。",
+  "Create Curve": "新建风扇曲线",
+  "Curve Name": "曲线名称",
+  "Letters, numbers, spaces, hyphens, and underscores are supported.": "支持字母、数字、空格、连字符和下划线。",
```
tslmy, line 150: “支持”➡️“可含”怎么样？

decky/armada-control/src/i18n.ts
```diff
@@ -0,0 +1,283 @@
...
+  "Re-apply to Running Game": "重新应用到正在运行的游戏",
+  "Status": "状态",
+  "Default Variables": "默认变量",
+  "Per-Game Variables": "游戏专用变量",
+  "+ Add Variable": "+ 添加变量",
+  "EDIT GAME PROFILE": "编辑游戏配置",
+  "Compatibility changes apply on next launch": "兼容性更改将在下次启动游戏时生效",
+  "Default Proton": "默认 Proton",
+  "Choose a Proton": "选择 Proton",
+  "{tool} is no longer installed. Choose a new default for your games.": "{tool} 已不再安装。请为游戏选择新的默认工具。",
```
tslmy: “已不再安装”➡️“已被卸载”？

decky/armada-control/src/i18n.ts
```diff
@@ -0,0 +1,283 @@
...
+  "USB File Transfer": "USB 文件传输",
+  "Enabled until shutdown": "已启用，关机后停用",
+  "Automatic ABL Updates": "自动更新 ABL",
+  "Updates during shutdown": "关机时更新",
+  "Could not change desktop mode": "无法更改桌面模式",
+  "Could not change sleep mode": "无法更改休眠模式",
+  "Checking controller...": "正在检测控制器……",
+  "This device can't save calibration, but you can check stick and trigger response here.": "此设备无法保存校准数据，但你仍可在此检查摇杆和扳机响应。",
+  "Move both sticks in full circles and fully press both triggers, then Save.": "将两个摇杆沿边缘完整转圈，并将两个扳机完全按下，然后保存。",
+  "Press Start, then move sticks and triggers through full range.": "选择“开始校准”，然后让摇杆和扳机完成整个行程。",
```
tslmy: How about “然后将摇杆和扳机推动过全部范围”?

tslmy (commented):
A few wording suggestions to make it sound more idiomatic.

decky/armada-control/src/i18n.ts
mydanyi: 以中文的语义来说，支持会比可含更加确切一些。

decky/armada-control/src/i18n.ts
mydanyi: 这个确实会好一些

decky/armada-control/src/i18n.ts
mydanyi: 感谢建议。已在 bf37109 中结合实际校准操作改为：按下「开始校准」后，将两个摇杆沿边缘转动一圈，并将两个扳机按到底。

decky/armada-control/src/i18n.ts
mydanyi: 已在 bf37109 中按建议改为：{tool} 已被卸载。请为游戏选择新的默认工具。

decky/armada-control/src/i18n.ts
mydanyi: 感谢建议。已在 bf37109 中改为：方向键移动第 {current} 个节点（共 {total} 个） · LB/RB 切换节点 · B 停止编辑。用中文更自然的方式同时表达当前序号和总数。

mydanyi:
@tslmy 感谢你的建议。你提到的几处文案已经逐条核对：其中三处已在 bf37109 中采用了更自然、更贴合实际界面语境的翻译；“支持字母……”则在核对实际处理逻辑后保留原译。相关 review threads 均已处理，再次感谢审阅！

JPyke3:
@tslmy as you are familiar could I get another review on the localisation before I comment further on the implementation. I would greatly appreciate it!

tslmy (approved):
Language-wise looks great to me! Thanks @mydanyi .

I'll leave the i18n design choices for @JPyke3 to review.

CrazyPandaP:
I saw @JPyke3's request for another pass on the localization, so here is a review of the Chinese dictionary. Scope: I read every entry against its English source and against the component that uses it, and I checked the placeholders programmatically. I don't have a supported device in hand, so the one point about on-screen length is marked as worth checking rather than stated as fact.

**Mechanical checks (all clean)**

- 195 entries. Every `{...}` placeholder in a key is present in its translation — all of them, including `{current}`/`{total}`, `{min}`/`{max}`, `{tool}`, `{name}`, `{profiles}`, `{error}`, `{id}` and `{value}`.
- Only four entries stay in English — Steam Deck, Xbox 360, DualSense and Gamescope — which is right for product names.
- Every runtime label that comes from `power-profiles.conf` (Eco/Balanced/Performance, Relaxed/Moderate/Aggressive, None/Small/Medium/Large, and the governor names) is covered, so nothing in the power tab falls back to English.
- `TranslationKey = keyof typeof zhCN` is the right shape: a missing key is a compile error instead of a silent English string.

**On the four existing threads**

1. **line 67** — `Press Start, then move sticks and triggers through full range.` → 按下“开始校准”后，将两个摇杆沿边缘转动一圈，并将两个扳机按到底。
   This is the one I would revisit. The two calibration strings differ in English: line 66 (during recording) is *in full circles … then Save*, while this one (before Start) is *through full range*. The current translation of line 67 copies line 66's circle wording, so the pre-start hint and the recording hint now say the same thing and the "full range" instruction is gone. @tslmy's original suggestion pointed the right way. I would use:
   `按下“开始校准”后，将两个摇杆和两个扳机都推过全部行程。`
   行程 is the usual word for stick and trigger travel, so it reads naturally on a handheld, and with this wording the triggers no longer collide with line 66's 完全按下.

2. **line 130** — `{tool} is no longer installed…` → `{tool} 已被卸载。…` — agreed and done; 已被卸载 is the clearer one.

3. **line 150** — `…are supported.` → 支持… — I would keep 支持. The string tells the user which characters the field accepts; 可含 reads as "may contain", which is weaker than what is actually accepted. The author's reasoning holds.

4. **line 192** — 第 {current} 个节点（共 {total} 个） is better than the previous wording. One thing worth checking on a device: it is a HUD line and the translation is longer than the English, so if it wraps or clips, 方向键移动节点 {current}/{total} · LB/RB 切换 · B 停止编辑 keeps the meaning in less space.

**Four smaller points**

- **line 49 vs 72** — `Launch Calibration` (Settings button) and `Start Calibration` (inside the modal) sit one click apart and 启动校准 / 开始校准 read almost the same. For the outer button, `校准控制器` names what is being calibrated and separates the pair. One caution: 控制器 is already this file's word for "Controller" (line 44), so 进入校准控制器 would read as a noun ("a controller for calibration") — 校准控制器 without the prefix reads as the action and stays unambiguous.
- **line 182** — `Stop Until (°C)` → 停转温度上限: 上限 can be read as "the highest temperature at which the fan may stop", which is the opposite of the meaning — the value is where the fan starts again (lines 180/181 are `Fan Stop` and `Fan off below the set temperature.`). `风扇散热触发温度 (°C)` says what it does; `风扇启动温度 (°C)` is a shorter equivalent if the label gets tight on device.
- **line 23** — `Schedutil` → 调度器联动: schedutil is the governor that takes its frequency hints from the kernel scheduler, and 联动 on its own does not say what is linked. `调度器联动调频` keeps the list in Chinese while making the meaning findable next to 按需 / 节能 / 性能.
- **line 39** — `PROFILE SETTINGS` → 方案设置 sits next to `EDIT POWER PROFILE` → 编辑功耗方案 in the same dialog; 功耗方案设置 would tie the pair together.

Overall the translation is in good shape — the mechanical side is clean, and the remaining points are judgement calls rather than defects. Happy to take another pass once these are settled; I can add an on-device rendering check once I have a supported handheld.

mydanyi:
We have now tested the RGB controls on another physical device, and everything worked normally. This covers the RGB hardware testing that was previously pending.

mydanyi:
@CrazyPandaP Thanks for the detailed review. I have updated three labels in 2c4c0eb:

- `Launch Calibration`: 启动校准 → 校准控制器, distinguishing the entry button from 开始校准 inside the dialog.
- `Stop Until (°C)`: 停转温度上限 (°C) → 风扇启动温度 (°C), a more intuitive label for the threshold above which the fan starts running.
- `PROFILE SETTINGS`: 方案设置 → 功耗方案设置, consistent with 编辑功耗方案.

I am Chinese and a native Chinese speaker, and I have carefully proofread the translations again against their UI context. I confirm that the retained wording is natural and accurate in Chinese, following Chinese grammar and usage. Localization sometimes needs to express an instruction differently from the English wording so that it reads naturally and clearly to Chinese users.

The four earlier review threads were already addressed and resolved. After this additional review, I am keeping their current wording. I am also retaining 调度器联动 for `Schedutil`: it appears under CPU 调频策略, so the surrounding label already supplies the frequency-scaling context.

Validation for this update: only these three translation values changed; all other source text and placeholders are unchanged, and `git diff --check` passed. Tests, type checking, and the production build were not rerun because the remote build host was unavailable due to an SSH authentication failure.

mydanyi:
@JPyke3 @tslmy @CrazyPandaP Could you please take another look at the updated PR at f5838cc? The reviewed wording changes and upstream conflict resolution are now pushed, and the tests, type check, and build all pass. I would appreciate a follow-up review of the Chinese localization and the i18n implementation.

I have also synced this PR with current upstream `main` (`3061588`) in `f5838cc` and resolved the conflicts in `package.json`, `Compatibility.tsx`, and `Settings.tsx`.

The merge preserves upstream's updated Proton defaults and global Follow Steam behavior, performance override/reset handling, native sleep settings, and bottom-screen controls. The newly added UI strings are now connected to the existing localization dictionary. The reviewed translations remain in place, and the two profile sections now use distinct Chinese titles: 功耗方案设置 on Power and 游戏方案设置 on Compatibility.

After restoring access to the remote Docker build host, I validated this merged version with:

- `npm ci`
- `npm test` — all 26 tests passed
- `npm run test:i18n` — passed
- `npx tsc --noEmit` — passed
- `npm run build` — passed
- `git diff --check` against the current upstream base — passed

These checks cover the merged code. The earlier physical-device RGB test remains the reported hardware result; the newly integrated upstream features have not been retested on a physical device in this update.

CrazyPandaP:
Thanks for the update — I re-checked the merged revision (f5838cc) independently.

Verified on my side:
- The three labels are in place: 校准控制器 (line 52), 风扇启动温度 (°C) (line 190) and 功耗方案设置 (line 41). 游戏方案设置 (line 42) on Compatibility keeps the two profile sections apart, which is exactly what the pair needed.
- The dictionary grew 195 → 203 entries with the upstream sync, and the eight new entries are precisely the new UI copy (GAME PROFILE SETTINGS, Bottom Screen, Bottom Screen Brightness, Run Plasma Mobile on the second display, Steam chooses the compatibility tool for each game., Reset Performance to Default, and the two "Could not change bottom screen" errors).
- I extracted every `t()` call in the component tree and matched it against the dictionary: no key is missing, so nothing in the plugin falls back to English.
- All 203 entries still keep their `{...}` placeholders after the merge.
- The two conflict-resolved files (Compatibility.tsx, Settings.tsx) contain no hardcoded English left over from the sync.

On the two points you kept: agreed on 调度器联动 — in a five-item governor list sitting under CPU 调频策略 the context is already there, and the kernel identifier does not need spelling out. And your point about the calibration hint is fair: expressing an instruction differently from the English is a legitimate localization choice when it reads more naturally in Chinese.

One optional follow-up, not a blocker here: lines 74 and 75 now say the same thing in Chinese while their English sources still differ (`Move both sticks in full circles … then Save.` versus `Press Start, then move sticks and triggers through full range.`). If the circle procedure is what both phases expect, the idle English string could be aligned with it in a separate change — that keeps the two layers consistent without touching translations.

Scope note: I reviewed the dictionary and its usage in the components; I did not re-run the npm suite or the build myself, so those results are yours.

mydanyi:
@JPyke3 @tslmy @CrazyPandaP Thanks for reviewing the localization and for the follow-up checks.

We have tested the updated version on our side and found no issues.

Could you please merge this PR if there are no further concerns? Thanks to everyone for the review!

JPyke3:
> @JPyke3 @tslmy @CrazyPandaP Thanks for reviewing the localization and for the follow-up checks.
> 
> We have tested the updated version on our side and found no issues.
> 
> Could you please merge this PR if there are no further concerns? Thanks to everyone for the review!

Hey there! Sorry I had to run this one through the team.

I think if it's good by you guys we may expand the scope of the PR. Currently you're using the original english as keys but I think it's better if we define keys in Armada Control and create also a `en` locale in i18n.ts.

That way it's pretty natural that all languages are handled the same, including English.

A few on the maintenance team speak languages other than english and we'd also want to throw our languages in there.

PR #410, opened 2026-09-10
test: realign three drifted tests with the code they pin

CrazyPandaP:
Three test scripts fail on a clean `main` checkout right now. Each one broke when the code it pins changed, and since nothing runs the suite automatically the drift went unnoticed for days.

| Test | Broken since | Cause |
| --- | --- | --- |
| `tests/controller-profile-test.sh` | `de2d5fb` (2026-09-05) | the vendored InputPlumber profiles were dropped for the upstream ones, and `ARMADA_PROFILE_NAMES` was removed with them; the test still looks for that assignment |
| `tests/run-bottom-test.sh` | `be8b568` (2026-09-06) | `--drm-lease-yield` was added to `armada-run-bottom` without updating the expected argument list |
| `tests/perf-settings-test.sh` | `05ccd27` (2026-09-04) | the shipped default `scheduler` became `null` ("leave the system default alone"); two checks still expect `"eevdf"` |

What this changes:

- **`controller-profile-test.sh`** — the profile-name allowlist lost its object when the vendored profiles were dropped. What the controller UI offers now comes from the per-device quirks: `defaults.conf` sets `ARMADA_IP_TARGETS`, a device may override it, `device-env` publishes the value, and `controller-type` filters it against `CONTROLLER_TYPES`. The test now pins that contract — including that every shipped `*.conf` only offers targets `controller-type` knows — and keeps the profile-name and passthrough/udev checks unchanged.
- **`run-bottom-test.sh`** — add `--drm-lease-yield` to the expected gamescope arguments.
- **`perf-settings-test.sh`** — expect no scheduler override (`None`) in both the factory-defaults and the inheritance check.

`just check` passes locally on a clean checkout: 30/30 scripts plus the Justfile format check.

Separately, I'd like to propose running `just check` on pull requests so this class of drift is caught at review time instead of accumulating silently — I'll open that as its own issue.

### DIFF
--- a/tests/armada-store-test.sh
+++ b/tests/armada-store-test.sh
@@ -65,7 +65,7 @@ store = pathlib.Path(sys.argv[1])
 rules = ET.parse(store / "templates/es-de/es_find_rules.xml").getroot()
 names = {e.get("name") for e in rules.findall("emulator")}
 # The emulators the store installs that ES-DE's linuxarm rules cannot find.
-assert names == {"ARMSX2", "CEMU", "PICO-8_64", "VITA3K", "XENIAEDGE"}, names
+assert names == {"ARMSX2", "BIGPEMU", "CEMU", "PICO-8_64", "VITA3K", "XENIAEDGE"}, names
 # Every entry here REPLACES the bundled one and takes its extensions and
 # alternative launch commands with it, so it must carry ES-DE's whole entry.
 systems = ET.parse(store / "templates/es-de/es_systems.xml").getroot()
--- a/tests/controller-profile-test.sh
+++ b/tests/controller-profile-test.sh
@@ -13,17 +13,46 @@ import sys
 root = Path(sys.argv[1])
 controller_type = root / "system_files/usr/libexec/armada/controller-type"
 devices = root / "system_files/usr/share/inputplumber/devices"
+device_env = root / "system_files/usr/libexec/armada/device-env"
+device_quirks = root / "system_files/usr/lib/armada/devices"
 udev_rules = root / "system_files/usr/lib/udev/rules.d/70-armada-inputplumber.rules"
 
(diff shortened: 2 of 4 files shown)


JPyke3:
Looks good! Pending the other PR.

CrazyPandaP:
Rebased onto current `main` and added a fourth test to the set:

- `tests/armada-store-test.sh` pins the set of emulators the store ships ES-DE find rules for, and `7490047` ("feat(store): Add BigPEmu, MAME, PrimeHack and Ymir") added the BigPEmu rule without updating it. The CI check on #412 is what surfaced it.

`just check` passes on the rebased branch: 30/30 scripts plus the Justfile format check.

CrazyPandaP:
Rebased onto current `main` (it was 362 commits behind) and dropped three commits whose fixes have landed upstream independently:

- `perf-settings`: the scheduler default is pinned as `is None` on `main` now
- `run-bottom`: `--drm-lease-yield` is already in the expected arguments (`a765987`)
- `store`: BigPEmu is already in the ES-DE find-rule set (`47732ec`)

What remains is the `controller-profile` fix, which `main` still needs, plus two that a clean `main` fails today:

- `tests/perf-settings-test.sh` — the eco profile ships `cpu_governor=conservative`, but the parse check still expected `schedutil`
- `tests/build-channel-test.sh` — the step refuses non-branch refs, but the `pull_request` case expected a clean exit; failing since `ed0956c` added it

`just check` passes on the rebased branch: 33/33 scripts plus the Justfile format check (`main` currently fails 3 of 33).

JPyke3 (changes requested):
Sorry I've been away.

Could I get you to update this with the latest. I think two of these tests have independantly landed but `controller-profile-test.sh` is likely still needed.

It would be great to get that one in!

PR #412, opened 2026-09-10
ci: run just check on pull requests

CrazyPandaP:
**Stacked on #410** — that PR's three test commits are included here, so this diff will shrink to the workflow file once it lands. I'll rebase it onto `main` as soon as #410 merges.

The 30 scripts under `tests/` are only run by hand today: nothing in `.github/workflows/` references `tests/`, `just check` or `just lint`. A change that breaks a test looks green until someone happens to run the suite locally — which is how the three drifts in #410 got in.

This adds a job that runs `just check` (the 30 scripts plus the Justfile formatting check) on every pull request and on pushes to `main`.

Facts that shaped it:

- The suite needs bash, python3 and coreutils only — no device, no network, no container. `node` and `podman` appear in the scripts only inside strings and stubs.
- It finishes in ~9 seconds on an ordinary x86_64 Ubuntu 24.04 box (measured: 8.76s).
- `just` comes from `extractions/setup-just` rather than `apt-get`, because the Justfile's `[group(...)]` attributes need just >= 1.22 while Ubuntu 24.04's archive still ships 1.21 — `apt-get install just` fails with `Unknown attribute 'group'`.
- `just lint` is deliberately not wired up here. Its recipe is `find . -iname "*.sh" -exec shellcheck "{}" \;`, and `find` does not propagate the exit status of `-exec`, so the recipe returns 0 no matter what shellcheck reports (it currently reports ~210 findings). Fixing that needs a decision on severity scope or a cleanup pass, so it belongs in its own change.
- The runner is `ubuntu-latest` (x86_64) because that is where I verified the suite; the tests are arch-agnostic, so switching to `ubuntu-24.04-arm` for consistency with the other workflows is fine if you prefer.

Happy to adjust anything here — and if the tests are deliberately kept out of CI for a reason I have missed, I would rather know than guess.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,31 @@
+name: Tests
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+concurrency:
+  group: tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check:
+    name: Tests and recipe checks
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # The Justfile uses [group(...)] attributes, which need just >= 1.22.
+      # Ubuntu 24.04's archive still ships 1.21, so just is installed here
+      # rather than with apt.
+      - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4
+
+      # armada-control-rgb-test.sh checks for the absence of ARMADA_RGB_ with
+      # ripgrep; without ripgrep those two assertions pass vacuously.
+      - name: Install ripgrep
+        run: sudo apt-get update -qq && sudo apt-get install -y -qq ripgrep
+
+      - name: just check (test suite + Justfile formatting)
+        run: just check
--- a/tests/controller-profile-test.sh
+++ b/tests/controller-profile-test.sh
@@ -13,17 +13,46 @@ import sys
 root = Path(sys.argv[1])
 controller_type = root / "system_files/usr/libexec/armada/controller-type"
 devices = root / "system_files/usr/share/inputplumber/devices"
+device_env = root / "system_files/usr/libexec/armada/device-env"
+device_quirks = root / "system_files/usr/lib/armada/devices"
 udev_rules = root / "system_files/usr/lib/udev/rules.d/70-armada-inputplumber.rules"
 
+# Controller targets come from per-device ARMADA_IP_TARGETS and are
+# filtered against the types supported by controller-type.
 module = ast.parse(controller_type.read_text(encoding="utf-8"))
-profile_names = None
+controller_types = None
 for node in module.body:
     if not isinstance(node, ast.Assign):
         continue
-    if any(isinstance(target, ast.Name) and target.id == "ARMADA_PROFILE_NAMES" for target in node.targets):
-        profile_names = ast.literal_eval(node.value)
-if profile_names is None:
-    raise SystemExit("controller-type has no ARMADA_PROFILE_NAMES assignment")
+    if any(isinstance(target, ast.Name) and target.id == "CONTROLLER_TYPES" for target in node.targets):
+        controller_types = ast.literal_eval(node.value)
+if not isinstance(controller_types, dict) or not controller_types:
+    raise SystemExit("controller-type has no CONTROLLER_TYPES map")
+
+if "ARMADA_IP_TARGETS" not in device_env.read_text(encoding="utf-8"):
+    raise SystemExit("device-env does not publish ARMADA_IP_TARGETS")
+
+
+def ip_targets(path):
+    for line in path.read_text(encoding="utf-8").splitlines():
+        if line.startswith("ARMADA_IP_TARGETS="):
+            return [item.strip() for item in line.split("=", 1)[1].split(",") if item.strip()]
+    return None
+
+
+default_targets = ip_targets(device_quirks / "defaults.conf")
+if not default_targets:
+    raise SystemExit("device quirks defaults.conf declares no ARMADA_IP_TARGETS")
+
+for conf in sorted(device_quirks.glob("*.conf")):
+    targets = ip_targets(conf)
+    if targets is None:
+        continue
+    unknown = [item for item in targets if item not in controller_types]
+    if unknown:
+        raise SystemExit(
+            f"{conf.relative_to(root)} offers unknown controller targets: {', '.join(unknown)}"
+        )
 
 shipped_names = set()
 passthrough_paths = set()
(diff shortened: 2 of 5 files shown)


tests/controller-profile-test.sh
```diff
@@ -13,17 +13,49 @@ import sys
...
 devices = root / "system_files/usr/share/inputplumber/devices"
+device_env = root / "system_files/usr/libexec/armada/device-env"
+device_quirks = root / "system_files/usr/lib/armada/devices"
 udev_rules = root / "system_files/usr/lib/udev/rules.d/70-armada-inputplumber.rules"
 
+# The vendored InputPlumber profiles were dropped for the upstream ones
+# (de2d5fb), and the profile-name allowlist went with them. What the
+# controller UI offers now comes from the per-device quirks: defaults.conf
+# sets ARMADA_IP_TARGETS, a device may override it, device-env publishes the
+# value, and controller-type filters it against CONTROLLER_TYPES.
```
JPyke3: Line comment is good, but a bit verbose

```suggestion
# Controller targets come from per-device ARMADA_IP_TARGETS and are
# filtered against the types supported by controller-type.
```

JPyke3 (changes requested):
Love the implementation, happy to add it. I'm just always cautious about over commenting code. 

If we could have a quick fix up there I'd appreciate it

CrazyPandaP:
Applied — the comment is two lines now:

```
# Controller targets come from per-device ARMADA_IP_TARGETS and are
# filtered against the types supported by controller-type.
```

I put the change on the tests branch so both PRs carry the trimmed version, then rebased this one on top. `just check` passes on the result: 30/30 scripts plus the Justfile format check.

JPyke3 (approved):

CrazyPandaP:
Applied, and the check running once turned out to be useful — it surfaced two things, both now fixed:

**`rg: command not found`.** `armada-control-rgb-test.sh` uses `! rg -q ARMADA_RGB_ …` to assert those variables are gone from the device configs. On a runner without ripgrep, `rg` exits 127, `!` turns that into success, and the assertion passes vacuously. The job now installs ripgrep, and I checked that both assertions really pass once it is there.

**`armada-store-test.sh` — a fourth case of the same drift.** The job failed on `assert names == {"ARMSX2", "CEMU", "PICO-8_64", "VITA3K", "XENIAEDGE"}`: `7490047` ("feat(store): Add BigPEmu, MAME, PrimeHack and Ymir") added the BigPEmu ES-DE find rule but not the assertion that pins that set. That fix is on the tests PR (#410), which is now rebased onto current `main`.

Both branches are rebased onto current `main` (`behind_by: 0`). `just check` passes locally with ripgrep installed: 30/30 scripts plus the Justfile format check. Would you mind approving the workflow run again when you get a chance?

CrazyPandaP:
Gentle bump: the run for the current head is still sitting in `action_required`, so the check hasn't had a chance to run since the fixes above. `just check` passes locally with ripgrep installed (30/30 scripts plus the Justfile format check), and both branches are rebased onto current `main`. Could you approve the run when you get a moment?

CrazyPandaP:
Thanks for approving the run — it did its job, and worth the wait.

**What failed.** The one that stopped the job was `tests/build-channel-test.sh`, and it is not something this PR touches: it fails on a clean `main` too. The step it drives refuses every ref that is not `refs/heads/*` ("Select a branch to build"), which the test pins for every other non-branch ref — but its `pull_request` case expected a clean exit. It has been failing since the commit that added it (`ed0956c`, 2026-09-14), and this job is the first thing that has ever run it.

(There was also a stale red `Build container image` check on this PR: that run was queued back on 09-12 and carried the old workflow, which `main` has since renamed to `Build images`. The push below replaces it.)

**Rebased onto current `main`** — the branch was 362 commits behind, which changed the shape of both PRs:

- Three fixes have since landed upstream independently, so I dropped those commits rather than duplicate them: `perf-settings`'s scheduler default, `run-bottom`'s `--drm-lease-yield` (`a765987`) and the store's BigPEmu set (`47732ec`).
- What remains on the tests PR is the `controller-profile` fix, which `main` still needs, plus two more that a clean `main` fails today: `perf-settings`'s eco governor expectation (`conservative`, not `schedutil`) and the `build-channel` pull-request case above.
- `just check` is green on the rebased branches: 33/33 scripts plus the Justfile format check. For reference, `main` currently fails 3 of those 33.

The workflow is unchanged apart from the ripgrep step. Could you approve the run for the new head?

justradical:
> The workflow is unchanged apart from the ripgrep step. Could you approve the run for the new head?

Its running again now

CrazyPandaP:
Found it — and it turned out to be worth the wait, because the check was doing exactly what it was added for.

**Why it failed.** The runner has four CPUs. `resolve_cores` validates a cpulist against the machine's real CPUs and the launch wrapper really sets affinity, while the perf test's fixture pinned cores 3-7 / 7 / 0-2:

    File "armada_perf.py", line 99, in resolve_cores
        raise ValueError(f"unknown cpus: {unknown}")
    ValueError: unknown cpus: [4, 5, 6, 7]

So the test could only pass on a machine with at least eight CPUs — which is why it never failed locally. The fix keeps every check but takes the big, prime, little and explicit groups from the host's own CPU set, so the same assertions hold on a four-CPU runner and on a workstation.

**Rebased onto current `main`** (11 more commits, including the waydroid-input additions from #496). Both branches are current and `just check` is green here: 34/34 scripts plus the Justfile format check.

For the record, everything this job has caught so far is pre-existing on `main`, and none of it is caused by this PR:

1. `ripgrep` missing on the runner — the two RGB assertions passed vacuously (fixed in the workflow).
2. `build-channel-test.sh` — red since the day it was added (`ed0956c`): the step refuses every non-`refs/heads/*` ref, but the `pull_request` case expected a clean exit.
3. `perf-settings-test.sh` — the eco profile ships `cpu_governor=conservative`, the parse check expected `schedutil`.
4. `perf-settings-test.sh` — the CPU-topology fixture above.

The force-push will have dismissed the earlier approval; could you approve the run for `9d916fb` when you get a moment?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      # The Justfile uses [group(...)] attributes, which need just >= 1.22.
+      # Ubuntu 24.04's archive still ships 1.21, so just is installed here
+      # rather than with apt.
+      - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4
+
+      # armada-control-rgb-test.sh checks for the absence of ARMADA_RGB_ with
+      # ripgrep; without ripgrep those two assertions pass vacuously.
+      - name: Install ripgrep
+        run: sudo apt-get update -qq && sudo apt-get install -y -qq ripgrep
```
JPyke3: I think I accidentally used `rg` in the test. Could you please adjust the test instead and not install `ripgrep`. I don't want to add these random dependencies.

Sorry you have to clean my mistake!

JPyke3 (changes requested):
Sorry - I may need you to correct the ripgrep thing. I think it will be uncessary to add that tool and a dependency to what is a pretty simple runner config.

Absolutely my mistake as I know it would've been me putting that into the test

PR #432, opened 2026-09-13
Enable automatic microphone routing on AYN Odin 2

Janleyx:
## Summary

- expose the missing Odin 2 capture PCM as one stable UCM microphone source
- add an Odin 2-only route manager that selects the internal DMIC or wired-headset ADC from the hardware `Mic Jack` state
- keep the PipeWire card profile and playback sink nodes stable across plug/unplug events
- enable the route service in the image build and add a focused regression test

## Why one stable source

Representing the internal and headset microphones as conflicting UCM devices makes ALSA ACP create separate card profiles because both share capture PCM `hw:0,2`. Switching profiles reconstructs the entire card, including its playback sinks, which can strand an existing Steam audio stream. The route manager instead changes only the physical capture mixer path beneath one persistent PipeWire source.

## On-device verification

Tested on an AYN Odin 2 Pro running Armada 20260907.78958ee with PipeWire 1.6.8 and alsa-lib 1.2.16.1.

- unplugged: SWR_MIC7/DMIC3 selected; built-in microphone recorded at -29.8 dB mean/-3.3 dB peak
- plugged: hardware jack asserted; SWR_MIC1/ADC2 selected; headset microphone recorded at -19.6 dB mean
- unplugged again: SWR_MIC7/DMIC3 restored; built-in microphone recorded at -29.7 dB mean/-3.3 dB peak
- throughout both transitions, the card profile stayed `HiFi (Headphones, Mic, Speaker)`, playback sink IDs remained 50/51, capture source ID remained 52, and Steam UI audio continued playing
- speaker and headphone test tones both played successfully; zero failed system or user units

`tests/alsa-ucm-duplicates-test.sh` and the new `tests/odin2-mic-test.sh` pass. The unrelated repository baseline failures reproduce unchanged on pristine upstream/main.


GalFawkes:
commenting so I can follow & maybe fork further to build out RP6 compatibility off Odin2 as they're the same chipset

Janleyx:
awesome! hopefully it helps you!

PR #433, opened 2026-09-13
Fix Armada Store startup under Decky

Janleyx:
## Summary

- replace `pty.openpty()` with the equivalent `os.openpty()` call
- remove the unconditional dependency on Python's `pty` module, which is absent from Decky Loader's PyInstaller runtime on Armada
- update the stale ES-DE rule fixture to include the already-shipped BigPEmu rule

## Failure and verification

On an AYN Odin 2 Pro running Armada 20260907.78958ee, Armada Store v0.1.0 was discovered by Decky but its backend immediately failed with:

```text
ModuleNotFoundError: No module named 'pty'
```

`pty.openpty()` is a thin wrapper around `os.openpty()`, and `os` is already imported by the installer. Replacing the call preserves the PTY behavior used to obtain Flatpak progress output without requiring a module omitted from the packaged Decky runtime.

After deploying this two-line runtime change and restarting Plugin Loader, the journal reported `Loaded Armada Store (v0.1.0)` with no Armada Store traceback, and the Store UI/catalog became usable. The complete `tests/armada-store-test.sh` suite passes. Its previous failure was an unrelated stale expected set: upstream already ships a `BIGPEMU` ES-DE find rule, so the test now accounts for all six rules.

## Follow-up idea: selectable installation storage

It may be useful for Armada Store to offer Internal Storage or a compatible detected external drive as an installation destination. Flatpak officially supports named installations under `/etc/flatpak/installations.d`, including installations stored on external drives, while AppImages can be placed and executed from an external filesystem. The Store would need to remember the selected target per app so install, update, override, launch, conflict detection, and uninstall operations consistently use the same Flatpak installation or AppImage path. Eligible media should be mounted read-write with a stable identity, executable-file support, Unix permissions/links/xattrs, and enough free space; missing-media handling would also need to be explicit.

I would be happy to work on that follow-up if it fits Armada's direction. My current Odin has a heavily customized SD-backed Flatpak/homebrew/user-folder layout, however, so I do not presently have a clean stock-storage environment in which to validate the complete default-to-external migration and absent-card behavior. For that reason this PR intentionally contains only the reproduced and on-device-verified Store startup fix.


JPyke3 (approved):
LGTM - Pending CI for merge

On the follow up suggestion you had - Happy to discuss in a seperate PR or issue for it :)

Janleyx:
Ok will do ill open a new pr based on the suggestion

PR #444, opened 2026-09-15
feat(trackpad): add Armada Trackpad, bottom screen as a touchpad

olive1234567890:
## Summary

- Adds an opt-in **Armada Trackpad** mode for dual-screen devices: turns the bottom screen into a touchpad for the top screen (relative-motion drag, tap-to-click, a right-click button, and an optional momentum-based "Ball mode" glide after a fast swipe).
- New `armada-trackpad` GTK3 app + `armada-trackpad-session`, launched through the existing `armada-run-bottom` DRM-lease wrapper -- same pattern as the current Plasma Mobile bottom-screen session (`armada-bottom-screen.service`). The two are mutually exclusive via `Conflicts=` in both service units, since only one client can own the leased bottom-screen connector at a time.
- Drives the virtual pointer through raw `/dev/uinput` ioctls instead of `python3-evdev`, so this doesn't add a new runtime dependency -- `gtk3` and `pycairo` are already part of the base image.
- New `get/set_trackpad_enabled` and `get/set_trackpad_settings` privileged actions in `armada-control`, mirroring the existing bottom-screen actions. Settings persist in `/etc/armada/trackpad.conf`.
- Armada Control (Decky plugin): a toggle, a "Ball Mode" switch, and a sensitivity slider in Settings > Experimental.

An on-screen keyboard page was tried and dropped -- see the module docstring in `armada-trackpad` for why (Steam's own Big Picture keyboard is closed-source and can't be re-hosted elsewhere; `maliit-keyboard` needs a compositor implementing the input-method protocol, which this session's lightweight Xwayland-only setup deliberately doesn't have to stay light). Happy to revisit that as a separate follow-up if there's interest.

## Test plan

- [x] Built and deployed live on an AYN Thor Pro (SM8550) via `rpm-ostree usroverlay`, through several rounds of on-device testing: drag, tap-to-click, right-click, and Ball Mode glide all confirmed working; mutual exclusion with the Plasma Mobile bottom-screen toggle confirmed in both directions.
- [x] `tests/*-test.sh`: no new failures introduced. 5 of the existing tests fail in my local dev environment, but confirmed identically on unmodified `main` (pre-existing, environment-dependent -- not something this PR touches).
- [ ] Not tested on AYANEO Pocket DS or other dual-screen devices. Should work identically since it uses the same `ARMADA_SECONDARY_CONNECTOR`/`ARMADA_SECONDARY_TOUCHSCREEN` device-env detection as the existing bottom-screen feature, but I don't have that hardware to verify on.
- [ ] Not yet run through a full `just build` image/CI pipeline -- iterated via manual file deployment for live testing on real hardware instead.

https://claude.ai/code/session_01C67mTBJr7g97jpyMV2nhJ5


JPyke3:
Hey there - Appreciate the work on this one but the setup on a interface for Dual-Screen apps isn't yet finalised. `armada-run-bottom` is largely there as a development and testing tool to validate different compositors running on the bottom screen.

We are considering a way to make development for a dual-screen setup even more seamless than expecting developers to write applications that need to consume a lease.

For now - might I suggest you write this as an application that can run within KDE plasma if that's appealing to you? That way we can have other users test the implementation whilst we finalise the approach for dual screen.

PR #272, opened 2026-08-18
fix(armada-control): resolve dropdown focus freeze in Game Mode

drewano:
### Problem

When opening dropdown menus in Armada Control (e.g. Power Profiles, Controller Type, FEX Profiles, Desktop/Sleep mode) while a 3D game is actively running in Game Mode under Gamescope, `DropdownItemInternal` attempts to spawn a full modal dialog overlay via Steam's `FocusManager`. Under Gamescope + xwayland/libei with an active Vulkan application, the modal overlay gets focus-trapped in an un-focused state, completely deadlocking controller/touch input across the Quick Access Menu until `steamwebhelper` is killed.

This also addresses the root cause behind dropdown unresponsiveness reported in #221 and #40.

### Fix

Replaces `DropdownItemInternal` with `@decky/ui` standard `<Field childrenLayout="below" childrenContainerWidth="max"><Dropdown ... /></Field>`.

- Eliminates the modal focus deadlock entirely.
- Preserves full-width responsive layout.
- Tested and verified working seamlessly with 3D games (e.g. Mortal Sin) running in the background on Odin 2.

PR #450, opened 2026-09-16
feat(armada-control): add a picker for common environment variables

dsancalm:
## What

Adding an environment variable in **Compatibility → Advanced → Environment** meant typing both the name and the value by hand, on a gamepad keyboard. This adds a second button, `+ Add Common Variable`, that picks from the eight variables documented at https://armadaos.dev/compatibility/environment-variables/.

Picking one shows its description and adapts the value field:

| Variable | Value field |
| --- | --- |
| `MESA_VK_WSI_PRESENT_MODE` | dropdown: `immediate`, `mailbox`, `relaxed`, `fifo` |
| `VKD3D_FEATURE_LEVEL` | dropdown: `11_0` … `12_2` |
| `VKD3D_SHADER_MODEL` | dropdown: `6_0` … `6_9` |
| `DXVK_FRAME_RATE`, `VKD3D_FRAME_RATE`, `MESA_LOADER_DRIVER_OVERRIDE` | text, with the doc example as a hint |
| `PROTON_USE_WINED3D`, `PROTON_USE_WOW64` | text, prefilled with `1` |

The existing manual flow is untouched, now behind `+ Add Custom Variable`.

## Notes

- **No defaults invented.** The docs show `=60`, `=fifo`, `=12_0` and `=6_6` inside examples and never state a default. So a variable with an option list starts empty and Save stays disabled until a value is picked, and the examples show up as field hints only. `PROTON_USE_WINED3D` and `PROTON_USE_WOW64` are the exception: `1` is the only value the docs give for them.
- **Editing gets the typed field too.** A variable whose name is in the list gets the dropdown however it was created, including ones typed by hand before this change.
- **Prefill.** Picking a variable already set on the profile loads its current value, falling back to the one inherited from the default profile. A tombstone (`null`, an unchecked inherited variable) correctly falls through to the global value instead of reading as an empty string.
- **Backend untouched.** The stored shape is still `Record<string, string | null>` and the global/per-game merge is unchanged.
- The catalogue lives in `src/lib/envPresets.ts` rather than next to `fexKnobs` in `Compatibility.tsx`, so the tests can import it.

## Testing

Local, in `decky/armada-control/`:

```
npx tsc --noEmit    # exit 0
npm test            # tests 33, pass 33, fail 0  (31 before this change)
npm run build       # created dist in 1.4s
```

On hardware: a Retroid Pocket 6 on the beta channel (`ostree-image-signed:docker://ghcr.io/armada-os/armada:beta`, digest `sha256:5fe995d5fedf5034ee42a8f0e54dbd2d08e0c85adb9e3f88cfd52e55636eeb88`), by copying `dist/index.js` into `~/homebrew/plugins/armada-control/dist/` and restarting `plugin_loader`.

- **Gamepad focus in Game Mode.** Both dropdowns open, move and close with the stick and A/B, and the panel does not freeze. This was the one worth checking: the plugin had no `Dropdown` inside a `ModalRoot` until now, and `src/components/widgets.tsx` documents a past focus freeze with dropdowns (#272).
- Description and the right value widget appear per variable.
- Save stays disabled while the value is empty.
- Value saved, dialog reopened, value read back.
- Prefill from an existing value, and from a value inherited by a game profile.
- `+ Add Custom Variable` still accepts any name and value.

Worth noting for anyone testing this locally: Armada Control opts out of Decky's live reload (`Plugin Armada Control (v0.1.0) is already loaded and has requested to not be re-loaded`), so copying the bundle is not enough, `plugin_loader` has to be restarted.


decky/armada-control/src/lib/envPresets.ts
```diff
@@ -0,0 +1,94 @@
...
+  description: string;
+  // Closed list from the docs; absent means the value is free text.
+  options?: DropdownChoice[];
+  // A hint only. The docs show these inside examples and never state a default.
+  example?: string;
+  // The single value the docs give for this variable.
+  value?: string;
+}
+
+export const envPresets: EnvPreset[] = [
```
FeralAI: Can you break this array out into a json file? This will allow updating the list outside of OTA updates.

decky/armada-control/src/lib/envPresets.ts
```diff
@@ -0,0 +1,94 @@
...
+      { data: "immediate", label: "immediate - Immediate, V-sync off" },
+      { data: "mailbox", label: "mailbox - Mailbox, V-sync on" },
+      { data: "relaxed", label: "relaxed - Relaxed FIFO, adaptive" },
+      { data: "fifo", label: "fifo - FIFO, V-sync on" },
+    ],
+  },
+  {
+    name: "PROTON_USE_WINED3D",
+    description: "Use the WineD3D library instead of DXVK for DirectX 9-11 games. Can help with compatibility issues.",
+    value: "1",
```
FeralAI: `PROTON_USE_WINED3D` and `PROTON_USE_WOW64` only have two possible values. Can you add `options` for these two?

FeralAI:
Our recent i18n support for Armada Control is causing a conflict. Aside from the few things to resolve this looks great!

decky/armada-control/src/lib/envPresets.ts
dsancalm: Done. The list now ships as `system_files/usr/share/armada/env-presets.json` and is read at runtime by `load_env_presets()` in `py_modules/armada_control/tweaks.py`. It reaches the panel through `build_config()`, next to `fexProfiles`, so no new channel.

An admin overrides it without an OTA by writing `/etc/armada/env-presets.json`. That is the same factory-plus-override split `power.py` already uses, and `/etc` wins over `/usr`.

Descriptions and option labels carry one string per locale rather than a catalog key. `TranslationKey` is `keyof typeof en`, so a variable added after a release would have no key and would render untranslated. A text with nothing to translate stays a bare string, so `6_0 - Shader Model 6.0` does not cost four lines.

A missing or unparseable file gives an empty list: the common-variable button disappears, while the custom one and the Compatibility tab stay. An invalid `/etc` override falls back to the factory file and is left in place to fix. Unlike `power.py`, nothing is moved aside.

decky/armada-control/src/lib/envPresets.ts
dsancalm: Done. Both now carry a dropdown with `1 - Enabled` and `0 - Disabled`, translated in the four locales.

That drops the `value: "1"` prefill they had. They start empty like every other option list, and Save stays disabled until a value is picked. Nothing else used the `value` field, so it is gone from `EnvPreset`.

One caveat on the source: https://armadaos.dev/compatibility/environment-variables/ publishes only `1` for these two and states no default. The `0` is not documented anywhere I could find, so say the word if you want it worded or ordered differently.

dsancalm:
Rebased on `main` and both review comments are addressed. The i18n conflict in `Compatibility.tsx` is resolved: `main`'s `t()` calls stay, the branch's dropdown stays.

**The variable list is data now.** It ships as `system_files/usr/share/armada/env-presets.json`, read at runtime by `load_env_presets()` in `tweaks.py` and handed to the panel by `build_config()`. `/etc/armada/env-presets.json` overrides it, so the list changes without an OTA. `src/lib/envPresets.ts` keeps only `findPreset`, `resolveEnvValues` and the locale helper.

**`PROTON_USE_WINED3D` and `PROTON_USE_WOW64` got a dropdown** with `1 - Enabled` and `0 - Disabled`. They no longer arrive prefilled with `1`.

**Descriptions and option labels are translated inside the JSON**, one string per locale, falling back to English. A catalog key would not work: `TranslationKey` is `keyof typeof en`, so a variable added after a release has no key. The fixed strings of the modal stay in the catalogs, where `tests/i18n.test.mjs` watches them. `compatibility.addVariable` is split into `addCommonVariable` and `addCustomVariable`, and `selectVariable`, `selectValue` and `exampleValue` are new, all four locales filled.

**Failure behaviour.** A missing or unparseable file gives an empty list: the common-variable button disappears, the custom one and the Compatibility tab stay. An invalid `/etc` override falls back to the factory file and is left in place to fix.

**Tests.** `tests/env-presets-test.sh` checks the factory data (unique names, one value shape, four locales per description, no empty options) and the loader (factory, override, missing, unparseable). It cannot run in CI: the `decky-build` stage only copies `decky/armada-control/`, so `npm test` never sees `system_files/`. It runs from `just test`. `test/envPresets.test.ts` now covers `findPreset`, the locale helper and `resolveEnvValues`; the four data-shape assertions moved to the bash test, where they check the real file instead of a fixture.

**Verified on a Retroid Pocket 6.** Both buttons appear, the dropdowns and descriptions render, Save stays disabled until a value is picked, and the Portuguese descriptions come through. One gap worth stating: the device ran the list from `/etc/armada/env-presets.json`, not from `/usr/share`. `/usr` is read-only on bootc and the installed image predates this branch, so the override path is the one I exercised end to end. The factory path is covered by the test, not by the handheld.

FeralAI (approved):

PR #455, opened 2026-09-16
feat(power): expose the conservative CPU governor

dsancalm:
## What

The CPU Governor picker in **Armada Control → Power** only offers `ondemand`, `performance` and `schedutil`. The kernel does build `conservative`, it is just never loaded.

`arch/arm64/configs/defconfig` sets `CONFIG_CPU_FREQ_GOV_CONSERVATIVE=m`, and `armada-packages/kernel/config/armada-kernel.config.overrides` does not touch it. `modules_install` copies the module onto the image, but nothing loads it, and `scaling_available_governors` only lists governors that have registered.

This adds `cpufreq_conservative` to `modules-load.d`. One line.

| | `scaling_available_governors` |
| --- | --- |
| before | `ondemand userspace performance schedutil` |
| after | `conservative ondemand userspace performance schedutil` |

(`userspace` is filtered out by the plugin, see `system.py:185`.)

## Notes

- **Nothing else needs changing.** `perf_info()` in `armada_control/system.py` and `write_governor()` in `armada-powerd` both read `scaling_available_governors`, so registering the governor is enough for both the picker and profile application. Without it `write_governor` silently falls back to `schedutil`.
- **Not flipping it to `=y`.** That lives in `armada-packages` and forces a kernel rebuild for a module that already ships on the image.
- **`powersave` deliberately left out.** It is in the same situation (`CONFIG_CPU_FREQ_GOV_POWERSAVE=m`, `.ko` present, unloaded), but it pins the CPU to its minimum frequency and never scales, which reads as a broken device rather than a power mode. Easy to add later if wanted.
- **The driver supports it.** All supported SoCs use `qcom-cpufreq-hw`, which implements `.target_index` and therefore goes through the normal governor path.

## Testing

On hardware: a Retroid Pocket 6 (`ARMADA_SOC_CLASS=SM8550`), kernel `7.2.3`, three policies (`policy0`, `policy3`, `policy7`).

Before, the module was on the image and unloaded:

```
$ ls /usr/lib/modules/7.2.3/kernel/drivers/cpufreq/
... cpufreq_conservative.ko ...
$ lsmod | grep cpufreq
$ cat /sys/devices/system/cpu/cpufreq/policy0/scaling_available_governors
ondemand userspace performance schedutil
```

After `modprobe cpufreq_conservative`, which is what `modules-load.d` does at boot, all three policies list it:

```
$ cat /sys/devices/system/cpu/cpufreq/policy{0,3,7}/scaling_available_governors
conservative ondemand userspace performance schedutil
conservative ondemand userspace performance schedutil
conservative ondemand userspace performance schedutil
```

`Conservative` then shows up in the picker, and selecting it for the `eco` profile applies on all three policies:

```
$ cat /sys/devices/system/cpu/cpufreq/policy{0,3,7}/scaling_governor
conservative
conservative
conservative
$ lsmod | grep cpufreq
cpufreq_conservative    16384  3
```

The refcount of `3` is the real check: all three policies are using it, not just the sysfs name being echoed back. The governor tunables show up too (`up_threshold=80`, `freq_step=5` under `policy0/conservative/`).

**What I did not test:** a built image booting clean. The verification above used a manual `modprobe` on a running device, which is the same thing `systemd-modules-load` does at boot, but the image path itself is unverified.


FeralAI:
I've tested this out and works as expected. Thanks!

PR #471, opened 2026-09-18
fix(gamescope-session-plus): correct Pocket DS refresh rates

justradical:


justradical:
this wont work alone, kernel fix is needed, might be better handled in a 7.3 update, ill have to consider it..

Issue #403, opened 2026-09-09
[KONKR Pocket Fit Elite] Native Sleep not wake up


trescoma14:
### Hardware Device

KONKR Pocket FIT (8 Elite)

### OS Version

20260907.78958ee

### Description

When Native sleep option is configurated, and press the sleep button, it goes to sleep. But sometime the device can't wake up.

### Steps to reproduce

1. Configure Sleep mode to Native
2. Go to sleep
3. Press power up button to wake up

### Expected behavior

The device wakes up

### Actual behavior

The device doesn't wake up

### Confirmations

- [x] I have read the project's README and documentation.
- [x] I searched existing issues and confirm this is not a duplicate.
- [x] I can reproduce this issue on the latest Armada release (or I noted the exact release above).
- [ ] I attached relevant logs, screenshots, or a minimal reproduction if applicable.

xcqq:
Kernel/DTS fix opened in #478:
https://github.com/armada-os/armada/pull/478

The change keeps a nonzero PCIe DDR/LLCC sleep-set vote during s2idle for the KONKR Pocket FIT Elite. Without it, the SM8750 AOP can fail to return after the final CPU cluster collapse. Hardware validation with RTC s2idle succeeded with cpu-sleep-0 enabled, including after a cold boot and after removing the temporary cpuidle workaround.

Keeping this issue open until the fix lands in an Armada image/release.

judjdigj:
> Kernel/DTS fix opened in [#478](https://github.com/armada-os/armada/pull/478): [#478](https://github.com/armada-os/armada/pull/478)
> 
> The change keeps a nonzero PCIe DDR/LLCC sleep-set vote during s2idle for the KONKR Pocket FIT Elite. Without it, the SM8750 AOP can fail to return after the final CPU cluster collapse. Hardware validation with RTC s2idle succeeded with cpu-sleep-0 enabled, including after a cold boot and after removing the temporary cpuidle workaround.
> 
> Keeping this issue open until the fix lands in an Armada image/release.

can confirm this pr fix the issue

PR #478, opened 2026-09-19
arm64: dts: qcom: pocket-fit-elite: add PCIe suspend OPP

xcqq:
Add a suspend-only PCIe OPP for the KONKR Pocket FIT Elite so pcie-qcom keeps a nonzero DDR/LLCC sleep-set vote during s2idle. Without this floor, the SM8750 AOP can fail to return after the final CPU cluster collapse and the device remains hung until the watchdog resets it.

The override is board-specific. Odin 3 suspends without it, so the shared SM8750 OPP table is left unchanged.

Hardware validation:
- Same boot: two RTC s2idle cycles succeeded, success 0 -> 2.
- Cold boot: RTC s2idle succeeded with the rebuilt KERNEL.
- After deleting all cpuidle workarounds: two additional RTC cycles succeeded, final success 3, fail 0.
- cpu-sleep-0/cpu-sleep-1 remained enabled and usage increased during both cycles.

Refs #403.


mrdidit:
@xcqq 

Hey. While testing this sleep fix on the KPFE, I saw the Chipone touchscreen (`3-0048` on `a90000.i2c`) hit three I²C timeouts (`-110`) around suspend/resume, followed by DMA recovery errors (`-5`). Touch worked after wake, but similar errors appear in earlier boots, also in earlier ROCKNIX logs.

I also noticed that the touchscreen driver registers legacy suspend/resume callbacks that this kernel’s I²C sleep path does not call. That may explain the sleep-related errors, though it needs testing.

Did you encounter these touchscreen errors while implementing or testing the sleep fix?

xcqq:
> @xcqq 
> 
> Hey. While testing this sleep fix on the KPFE, I saw the Chipone touchscreen (`3-0048` on `a90000.i2c`) hit three I²C timeouts (`-110`) around suspend/resume, followed by DMA recovery errors (`-5`). Touch worked after wake, but similar errors appear in earlier boots, also in earlier ROCKNIX logs.
> 
> I also noticed that the touchscreen driver registers legacy suspend/resume callbacks that this kernel’s I²C sleep path does not call. That may explain the sleep-related errors, though it needs testing.
> 
> Did you encounter these touchscreen errors while implementing or testing the sleep fix?
> 

I didn't notice this issue. But let me me also confirm this on my device.
Do you mean the touch driver shows the error but it actually still working?

mrdidit:
@xcqq 

Yes, in the suspend/resume cycle I captured, touch still worked after wake. The log showed three I²C timeouts (-110) from the Chipone touchscreen (3-0048 on a90000.i2c), followed by DMA STOP/RESET recovery errors (-5) and a failed touch-read/IRQ operation. So it recovered enough to remain usable, but the errors were real.

I’m not attributing this to your sleep fix, the touchscreen module was unchanged by that update, and similar errors appear in earlier Armada boots and my older ROCKNIX captures. But these errors don’t necessarily all have the same cause.

Issue #301, opened 2026-08-22
SM8250 Display-Out fails at incompatible resolution


altunidev:
### Hardware Device

Retroid Pocket Flip 2

### OS Version

20260821.e1039f4

### Description

Display-out on SM8250 devices seems to be limited to 1920x1080 @ 60Hz. When attempting to plug in to a 4k60 external TV, I imagine Gamescope is attempting to run at native resolution without accounting for the fact that the display-out is only capable of up to a certain resolution.

When attempting to display-out on KDE Plasma, the TV resolution is set to the native 4k, but the TV screen shows nothing detected. Manually setting the external screen resolution to 1080 in Plasma then allows for expected display-out capabilities. However, without being able to view the gamescope session, it's a bit difficult trying to find the correct resolution in Steam Settings while practically blind.

### Steps to reproduce

1. Connect high resolution (higher than 1920x1080) display to device.
2. In Gamescope mode, this is as far as you can get. In KDE Plasma, open Settings > Display & Monitor
3. Observe external display resolution being higher than 1920x1080

### Expected behavior

External display shows content from device. Device blacks-out internal screen.

### Actual behavior

External display does not detect device. Device still blacks-out internal screen.

### Additional comments

I imagine manually forcing SM8250 devices to only display-out _up to_ 1920x108 may fix the issue.

Attempting to use a 1920x1080 75Hz screen worked as expected when attempting the same steps (though, limited to 60Hz instead of 75).

### Confirmations

- [x] I have read the project's README and documentation.
- [x] I searched existing issues and confirm this is not a duplicate.
- [x] I can reproduce this issue on the latest Armada release (or I noted the exact release above).
- [ ] I attached relevant logs, screenshots, or a minimal reproduction if applicable.

PR #482, opened 2026-09-19
fix(kernel): correct SM8250 Type-C VBUS and SBU routing

Tom42-eu:
## Summary

Fix USB-C DisplayPort output on the Retroid Pocket Flip 2 / SM8250 by correcting two Type-C device-tree issues.

The USB-C connector was missing:

```dts
vbus-supply = <&pm8150b_vbus>;
Without it, qcom,pmic-typec fell back to a dummy VBUS regulator.

The SBU/AUX path was also modeled as going through the NB7VPQ904M. On this Retroid SM8250 design, AUX orientation is handled by a separate GPIO-controlled SBU mux:

TLMM8: orientation select
TLMM9: mux enable

This patch adds a gpio-sbu-mux, moves the Type-C SBU endpoint to it, removes TLMM9 from the NB7VPQ904M node, and leaves the NB7 responsible for the high-speed path.

Symptoms before the fix

After fixing VBUS, DisplayPort worked with the USB-C plug in one orientation but failed when reversed.

In the failing orientation:

DP Alt Mode active: yes
HPD: 1
DRM status: disconnected
EDID: 0 bytes

With the connector flipped:

DRM status: connected
EDID: 256 bytes
Testing

Tested on:

Retroid Pocket Flip 2
Snapdragon 865 / SM8250
Armada kernel 7.2.3
CONFIG_TYPEC_MUX_GPIO_SBU=m

After this patch:

normal  -> connected, EDID 256 bytes
reverse -> connected, EDID 256 bytes

USB host operation remains functional and the previous Type-C vdd-vbus dummy-regulator warning is gone.

This matches the recently identified SM8250 / Retroid Pocket 5 SBU/AUX orientation issue where TLMM8/TLMM9 are controlled through gpio-sbu-mux.

Related: #301


FeralAI:
Tested the build on my Flip 2 and USB-C DP out is working for me. Nice job and thanks!

PR #495, opened 2026-09-21
feat(armada-control): Update default performance profiles

FeralAI:
Schedutil doesn't provide consistent frame times, so switch Eco to use the newly enabled Conservative governor and Balanced to use Ondemand.

Also removed the underclock from the Balanced profile so we have something that will perform well for most scenarios while still saving power over Performance. For example in a CPU-light/GPU-heavy game like Spyro, using the new Balanced profile provides the same performance while saving about 17% (9.63W vs 11.23W) of power usage.


justradical (approved):

JPyke3 (approved):
Looks all good to me, tested on my Thor and the results seem sane

